PR #94, opened 2022-01-31
Left-Align markdown notice on 5th event creation page.

Jonas-Sander:
Original issue from GitLab: https://gitlab.com/codingbrain/sharezone/sharezone-app/-/issues/1444

Original look:

image

Fixed version:
image_2022-01-31_20-53-51

image_2022-01-31_20-53-23

I added a little bit of padding - this is how it would look like without (i think a bit too much left):

image_2022-01-31_20-54-22


Jonas-Sander:
Do you think its better with padding or not? I'm not 100% sure :D

nilsreichardt (approved):
LGTM

Yes, looks better with Padding 👍

PR #115, opened 2022-02-04
Add enter shortcut for class creation dialog.

Jonas-Sander:
When entering the school class name inside the school class creation dialog and pressing "Enter" the school class will now be created.

Before this change the TextField just lost focus after pressing "Enter".

### DIFF
--- a/app/lib/groups/src/pages/school_class/school_class_create.dart
+++ b/app/lib/groups/src/pages/school_class/school_class_create.dart
@@ -6,52 +6,61 @@
 //
 // SPDX-License-Identifier: EUPL-1.2
 
-import 'package:bloc_base/bloc_base.dart';
 import 'package:bloc_provider/bloc_provider.dart';
 import 'package:flutter/material.dart';
 import 'package:sharezone/groups/src/pages/school_class/my_school_class_bloc.dart';
 import 'package:sharezone_widgets/widgets.dart';
 
-class SchoolClassCreateBloc extends BlocBase {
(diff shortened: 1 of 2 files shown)


app/lib/groups/src/pages/school_class/school_class_create.dart
```diff
@@ -6,64 +6,81 @@
...
       content: SingleChildScrollView(
         child: Column(
           crossAxisAlignment: CrossAxisAlignment.start,
           children: <Widget>[
             const SizedBox(height: 8),
-            _NameField(),
+            _NameField(
+              value: className,
+              onChanged: (newName) => setState(() => className = newName),
+              onSubmitted: (className) => _createClass(context, className),
```
Jonas-Sander, line 74: This is the change for the "Enter" behavior

Jonas-Sander:
Otherwise I did a little refactoring. Instead of using an extra "Bloc" (that isn't a Bloc) we just use the StatefulWidget directly to save the class name.  
The Bloc was used elsewhere though so I moved the Bloc to the other location it was used.

app/lib/groups/src/pages/school_class/school_class_create.dart
Jonas-Sander: (`onSubmitted`)

nilsreichardt (approved):
LGTM

But let's set up a web preview for every pull request, so we are able to quickly test these changes. I already started with #116.

PR #136, opened 2022-03-20
Delete unused code.

Jonas-Sander:


Jonas-Sander:
@nilsreichardt Can you take a look - is there anything we should keep in your opinion?   
I just deleted everything unused (except stuff in generated files) so we can discuss here the question above.     
(Also look at the "Show diff" code thats automatically hidden if possible). 

Personally I'm unsure if we should just delete unused code always in general (without any exception). Maybe there's some stuff which has a high probability of being used again and if we delete it would just mean that we recreate the code again in the future (as we probably forgot that we had it someday).   
Do you think this applies to any code deleted in this PR or should we just delete everything?

nilsreichardt (approved):
LGTM

In general, we should always delete unused code. But I think it makes sense to make excuses, like adding features to packages which currently are not used by the app or website but made be used in the feature.

PR #139, opened 2022-03-24
Enable soft line break for markdown.

nilsreichardt:
## Description
Since last summer, it's possible to enable soft line breaks (https://github.com/flutter/packages/pull/437). Therefore, no `\` is required anymore.

I think adding the support for soft line break is the better solution instead of `\` because it's completely intuitive and for nearly every user it's new that you can break a line with `\`.

Prevents also confusions like https://discord.com/channels/545668860859252746/616181222577668106/956087823692361768

| Before | After |
| - | - |
|Screenshot 2022-03-24 at 03 16 43|Screenshot 2022-03-24 at 03 16 17 |

Text:
```
Guten Tag,
was geht.
asdf


asdfasf


hallo\
safdfs
```


Jonas-Sander:
@nilsreichardt Bei den Termindetails ist das scheinbar noch nicht aktiviert

Jonas-Sander (approved):
See comment above.  
LGTM for everything else.

nilsreichardt:
> @nilsreichardt Bei den Termindetails ist das scheinbar noch nicht aktiviert

Good catch!

PR #140, opened 2022-04-08
Use Flutter version in CI from top level `env` in file.

nilsreichardt:


Jonas-Sander (approved):
LGTM 👍   

Do you know if there is some way to define the flutter version in another file (maybe top level) and use it inside GitHub Actions?  
In this way we could maybe one day use e.g. `fvm` and GitHub Actions and always be sure to use the same version (version file would be the single source of truth)

nilsreichardt:
Yes, it's possible. We could normally use `fvm` and then install Flutter not with an GitHub Action (`subosito/flutter-action@v1`). So the workflow would be:

1. Install Dart
2. Install `fvm`
3. Install Flutter via `fvm`. 

Therefore, we would have a SOT for our Flutter version.

The downside is that this would probably extend the CI duration by 1 min (probably more) for just installing Dart, `fvm`, etc.

Jonas-Sander:
> Yes, it's possible. We could normally use `fvm` and then install Flutter not with an GitHub Action (`subosito/flutter-action@v1`). So the workflow would be:
> 
> 1. Install Dart
> 2. Install `fvm`
> 3. Install Flutter via `fvm`. 
> 
> Therefore, we would have a SOT for our Flutter version.
> 
> The downside is that this would probably extend the CI duration by 1 min (probably more) for just installing Dart, `fvm`, etc.

Can't we just read in a file with the version and use the normal flutter-action with the specified version?

nilsreichardt:
> Can't we just read in a file with the version and use the normal flutter-action with the specified version?

I think this is not possible with GitHub Actions

Jonas-Sander:
> > Can't we just read in a file with the version and use the normal flutter-action with the specified version?
> 
> I think this is not possible with GitHub Actions

https://stackoverflow.com/questions/61919141/read-json-file-in-github-actions

Wouldn't this enable our use-case?

nilsreichardt:
Ah, of course it's possible. I was thinking to set Flutter version at top level env, where we can't run shell commands. But in the job, we can easily read out a file and export the Flutter version as environment variable :+1:

PR #142, opened 2022-04-08
Fix wrong path trigger for `deploy-web-app` CD

nilsreichardt:


nilsreichardt:
Yes, true. But this mistake should have been caught in a review as well. A reminder to have small PRs for better review quality :)

PR #158, opened 2022-05-04
Replace `ThemeBloc` with `ThemeSettings` class to change ThemeBrightness, textScalingFactor and VisualDensity.

Jonas-Sander:
Replaced `ThemeBloc` with a new class called `ThemeSettings` that can be used to:
* Toggle between dark-mode, light-mode and using the system setting
* Change the `VisualDensity` of the app
* Change the `textScalingFactor`

Currently the user can still only toggle the brightness (e.g. dark-mode).  
Another PR in the Future might add a nicer UI to change also the other settings.

The `ThemeSettings` class is used via `Provider` and doesn't need a `StreamProvider`:
```dart
 @override
   // Will automatically rebuild if properties of [ThemeSettings] change.
   // See also `context.select` to only rebuild if the brightness changes.
   final themeSettings = context.watch<ThemeSettings>();

   return Column(
     children: [
       Text('Current brightness: ${themeSettings.themeBrightness}'),
       ElevatedButton(
         child: Text('Change to dark-mode'),
         onPressed: () => themeSettings.themeBrightness = ThemeBrightness.dark,
       )
     ],
   );
 }
```


nilsreichardt (approved):
Nice! 🚀 
LGTM

Can you add tests for `ThemeSettings`?

And we should also have Analytics, when someone changes the theme. Should we do this in another PR?

Issue #143, opened 2022-04-08
We should commit the `pubspec.lock` for library packages.


nilsreichardt:
~See: https://dart.dev/guides/libraries/private-files#pubspeclock~

~Our library packages are located in `lib`.~

Edit: Renaming this from "We should not commit..." to "We should commit". Not committing a mono repo where a library package is _only_ used in the repository doesn't really make sense to not commit the `pubpsec.lock`. Not committing the `pubspec.lock` would result into some problems like every developer which have a different dependency tree and maybe is the app not working for this person.

Jonas-Sander:
Ah didn't know that! Good to know, thanks ^^

nilsreichardt:
I think it makes to implement this ticket before starting with the null safety migration. Otherwise, it could be a nightmare with the dependencies.

Added this ticket to the "Migrate to null safety" milestone.

nilsreichardt:
I think this can be closed with pub workspaces.

PR #160, opened 2022-05-04
Run `sz pub get` with Flutter 2.5.2

nilsreichardt:
We had some `pubspec.lock` files in our `main` which were generated with > `2.5.2`.

### DIFF
--- a/app/pubspec.lock
+++ b/app/pubspec.lock
@@ -91,7 +91,7 @@ packages:
       name: archive
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "3.1.6"
+    version: "3.1.2"
   args:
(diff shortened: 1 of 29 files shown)


Jonas-Sander (approved):

nilsreichardt:
@Jonas-Sander Would be good to catch something like this in our CI, right? Should be relative easy. Just after running `analyze` (which also ran `sz pub get`) we check if there are git changes:
```sh
git add *
if ! git diff --cached --quiet
then
  echo "Some files are out sync with git: $(git diff)"
  exit 1
fi
```

nilsreichardt:
I think so, but maybe is not relevant anymore, when we implemented: #143

PR #153, opened 2022-04-24
Upgrade to Flutter v2.10.5

nilsreichardt:
* Changed version in `.fvm/fvm_config.json`
* Fix all analyzer warnings

## ToDo
* [x] Check builds for all platforms

PR #164, opened 2022-05-05
Replace `jitsi_meet` package with `jitsi_meet_wrapper` + increase iOS deployment target to 12.0

nilsreichardt:
## Description
I'm sure that the `jitsi_meet` package is the reason why the iOS build is failing in #153. The problem is that the `jitsi_meet` package isn't maintained anymore. So it's difficult to find a fix so that #153 can be merged.

There are also several issues that Android 12 is not supported anymore by the package (https://github.com/gunschu/jitsi_meet/issues/376, https://github.com/gunschu/jitsi_meet/issues/368, https://github.com/gunschu/jitsi_meet/issues/366).

There is a folk of `jitsi_meet` called `jitsi_meet_wrapper` which has better code and is maintained. So I migrated to this package (thanks to @saibotma 💙).

One big disadvantage: `jitsi_meet_wrapper` uses the current Jitsi iOS SDK which requires +12.0. This means we need to drop the support for iOS 11.0 (other option would be to remove the Jitsi Feature for all users) 😞

## Demo
| iOS | Android |
|-|-|
| image | image |

## ToDo
* [x] Run on Android emulator


Jonas-Sander (approved):

nilsreichardt:
Android is working :) I added a screenshot to the description.

PR #170, opened 2022-05-07
Add `coverage` folder to `.gitignore`

nilsreichardt:
I just saw that we had a `app/coverage` folder. I think this was a mistake to commit this folder. So I added `coverage/` to the top level `.gitignore`. Maybe Flutter also adds the `coverage` as default to the `.gitignore` file: https://github.com/flutter/flutter/issues/101349.


Jonas-Sander (approved):

Jonas-Sander:
Just for my understanding: Since it's top-level it will also apply for e.g. packages inside `/lib`, correct?

nilsreichardt:
> Just for my understanding: Since it's top-level it will also apply for e.g. packages inside `/lib`, correct?

Correct

PR #175, opened 2022-05-11
First integration test: Sign in with email address and password

### LINKED ISSUE #180: Set up integration tests
## Description
We should set up the integration tests in order to be able to add easily more integration tests.

nilsreichardt:
## Description
This PR introduces the first integration test. It's a simple one: Sign in with the email address and the password. The main focus of the PR was to set up the integration and fix the errors. Therefore, we can add more integration tests in different PRs.

## Demo
https://user-images.githubusercontent.com/24459435/167846068-8747bed6-f775-41f5-8703-c7e3a53a40ec.mov

## How to run the integration tests
```sh
flutter test \
  integration_test \
  --flavor dev \
  --dart-define \
  USER_1_EMAIL="YOUR@EMAIL.COM" \
  --dart-define \
  USER_1_PASSWORD="YOUR_PASSWORD"
```

## Native setup for integration tests
I skipped the steps for setting up the [iOS](https://github.com/flutter/flutter/tree/master/packages/integration_test#android-device-testing) and [Android](https://github.com/flutter/flutter/tree/master/packages/integration_test#ios-device-testing) device tests, because I think we only need these setups when run our tests with Firebase Test Lab or similar services. Firebase Test Lab needs these steps to properly run the tests. But at the moment, we just run our integration tests with `flutter test` in emulators/simulators.

## Related Tickets
Closes #180


app/integration_test/README.md
```diff
@@ -1,211 +1,53 @@
...
-  --target=integration_test/foo_test.dart \
-  --flavor prod
-```
-
-Der `foo_test.dart` **soll extra fehlschlagen** (er beinhaltet einen erfolgreichen und fehlschlagen Test-Fall).
-
-<details>
-  <summary>Beispiel-Output (hier wird mehrmals das gleiche ausgeprintet warum auch immer)</summary>
-  
-```
```
nilsreichardt, line 28: @Jonas-Sander I removed the old `README.md` for the integration tests because they were only just the documentation of how to set up the integration tests which is already good documented in the `README.md` of the `integration_test` package: https://github.com/flutter/flutter/tree/master/packages/integration_test

app/integration_test/README.md
```diff
@@ -1,211 +1,53 @@
...
+org-dartlang-app:/app_test.dart:34:30: Error: Non-nullable variable 'dependencies' must be assigned before it can be used.
+          beitrittsversuche: dependencies.beitrittsversuche,
+                             ^^^^^^^^^^^^
+org-dartlang-app:/app_test.dart:35:29: Error: Non-nullable variable 'dependencies' must be assigned before it can be used.
+          blocDependencies: dependencies.blocDependencies,
+                            ^^^^^^^^^^^^
+org-dartlang-app:/app_test.dart:36:28: Error: Non-nullable variable 'dependencies' must be assigned before it can be used.
+          dynamicLinkBloc: dependencies.dynamicLinkBloc,
+                           ^^^^^^^^^^^^
+```
```
Jonas-Sander, line 36: It seems like this might happen because on web it runs with null-safety? 

Couldn't we either:    

1. Use `late` for group variables
```dart
  late AppDependencies dependencies;
  late _UserCredentials user1;
```

2. Pass `--no-sound-null-safety` explicitly

3. Make the `app_test` file use null-safety by using `// @dart=2.14`

(I haven't tried any of it - just general ideas.)

app/integration_test/app_test.dart
```diff
@@ -0,0 +1,88 @@
...
+    // Credentials are passed via environment variables. See "README.md" how to
+    // pass the them correctly.
+    user1 = _UserCredentials(
+      email: const String.fromEnvironment('USER_1_EMAIL'),
+      password: const String.fromEnvironment('USER_1_PASSWORD'),
+    );
+
+    // We should ensure that the user is logged out before running a test, to
+    // have fresh start.
+    await dependencies.blocDependencies.auth.signOut();
```
Jonas-Sander, line 35: For the future we might make our own test api `IntegrationTestFixture` for common behaviors like e.g. logging out so that we don't couple our tests too much to the implementation.   

`await testFixture.auth.logOut();` (`auth` would probably also be our own class, not `FirebaseAuth`?) Idk. Let's see when we implement further tests.

app/integration_test/README.md
```diff
@@ -1,211 +1,53 @@
...
-end of failure 1
-
-
-
-Stopping application instance.
-Driver tests failed: 1
+### Mobile
+You can run the integration tests using the `flutter drive` command:
+```sh
+flutter test \
```
Jonas-Sander, line 12: Shouldn't we use `fvm flutter` for the examples?

app/integration_test/app_test.dart
```diff
@@ -0,0 +1,88 @@
...
+import 'package:sharezone/main/sharezone.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  AppDependencies dependencies;
+  _UserCredentials user1;
+
+  setUpAll(() async {
+    dependencies = await initializeFlutterApp(isIntegrationTest: true);
```
Jonas-Sander, line 22: No problem for now but does it fundamentally even make sense that the app **returns** dependencies? Wouldn't we pass dependencies to the app to run?   

I mean it might make sense to pass dependencies and still have a way to access the objects created with those dependencies, I guess? Idk. 

Like I said it's okay for now, but it kinda confuses me.

app/lib/main/run_app.dart
```diff
@@ -106,57 +125,69 @@ Future runFlutterApp() async {
...
+  // documents which results into a "[cloud_firestore/permission-denied] The
+  // caller does not have permission to execute the specified operation."
+  // because the user isn't signed anymore.
+  //
+  // A good solution would be to refactor the listeners. But as a quick
+  // workaround we just skip the listeners for integration tests.
+  if (!isIntegrationTest) {
+    listenToAuthStateChanged().listen((currentUser) {
+      if (currentUser?.uid != null) {
+        final sharezoneGateway = SharezoneGateway(
```
Jonas-Sander, line 142: Why is `listeners` plural? Do you mean that the `StreamSubscription` is not disposed of correctly? Or a you talking about the Firestore listeners?   

I don't fully understand the explanation.   

Is the following explanation correct?    

The code below that uses `listenToAuthStateChanged` creates `SharezoneGateway` which automatically subscribes to document changes in several collections in Firestore.   
This results in errors if the user logs out since we don't correctly dispose or close the streams inside `SharezoneGateway`.  The errors will cause any integration test to automatically fail.

Currently we just work around that by not executing the code below at all (and thus not creating `SharezoneGateway`). In the future we should ensure that the `SharezoneGateway` does not unnecessarily subscribe to document changes and that we can properly dispose of the subscriptions before logging out.

--------

I still don't fully get why the `  if (!isIntegrationTest) {` fixes it. Don't we also create `SharezoneGateway` for the normal usage of the application? Wouldn't this still lead to the same problems?

Jonas-Sander (changes requested):

app/integration_test/README.md
nilsreichardt: I also tried this, but you will get a different error other packages require null safety.

app/integration_test/app_test.dart
nilsreichardt: Yeah, we can definitely rework this, but it requires some refactoring in the `main.dart` because we do so many dirty things in the `runFlutterApp` method 😅 Therefore, my plan is tough in the app the thing which are blocking the integration tests, write a lot of integration tests and then start refactoring.

app/lib/main/run_app.dart
nilsreichardt: Discussed with Jonas offline the answers.

app/integration_test/README.md
nilsreichardt: Yes makes more sense when we use `fvm`

nilsreichardt:
@Jonas-Sander Can you re-review?

app/lib/main/run_app.dart
```diff
@@ -77,7 +77,10 @@ Future<void> runFlutterApp() async {
   );
 }
 
-Future<AppDependencies> initializeFlutterApp({
+/// Initializes the dependencies of the Flutter app.
+///
+/// Returns the initialized dependencies.
+Future<AppDependencies> initializeDependencies({
```
Jonas-Sander, line 83: To be honest I think this documentation is unnecessary. It doesn't give any more information than which the name and return type of the method already gives.  

Either delete the Docs or add something which isn't conveyed by the method itself. E.g. why the dependencies are initialized in a seperate method (so that the integration test can use this method) or whatever.

```suggestion
Future<AppDependencies> initializeDependencies({
```

app/lib/main/run_app.dart
```diff
@@ -128,14 +131,22 @@ Future<AppDependencies> initializeFlutterApp({
...
-  // workaround we just skip the listeners for integration tests.
+  // Currently we just work around that by not executing the code below at all
+  // (and thus not creating SharezoneGateway). In the future we should ensure
+  // that the SharezoneGateway does not unnecessarily subscribe to document
+  // changes and that we can properly dispose of the subscriptions before
+  // logging out. The listeners of "listenToAuthStateChanged().listen()" are not
+  // disposed. Therefore, when signing out, the app still tries to stream the
+  // Firestore documents which results into a
+  // "[cloud_firestore/permission-denied] The caller does not have permission to
+  // execute the specified operation." because the user isn't signed anymore.
```
Jonas-Sander, line 149: Better but there is still missing that the UserGateway created inside the SharezoneApp widget tree is fine and we don't know why.     

Also the text seems a bit duplicated and confusing between the first (further up top) and this second paragraph.   

It's okay for now, if you want maybe try to make it a bit nicer but it's okay like this as well I suppose.

Jonas-Sander (approved):
Some very small nits, but LGTM. You can merge this if you want.

app/lib/main/run_app.dart
nilsreichardt: Yes, it really does not add any value xD I removed the docs.

app/lib/main/run_app.dart
```diff
@@ -117,46 +137,60 @@ Future runFlutterApp() async {
...
+      // because we would try to access the user data after the user signed out.
+      // This would result an instant fail of the integration tests.
+
+      if (userGateway != null) {
+        await userGateway.dispose();
+      }
+
+      if (sharezoneGateway != null) {
+        await sharezoneGateway.dispose();
+      }
```
nilsreichardt, line 171: @Jonas-Sander I have managed to modify `SharezoneGateway`, `UserGateway`, `ConnectionsGateway`, `DataDocumentPackage` and `DataCombinedPackage` a bit so that we are able to dispose them correctly (not for the whole app but for the `listenToAuthStateChanged` listener). Therefore, we can remove is `isIntegrationTest` flag 🙂

nilsreichardt: But i need to check manually that this doesn't break anything in the app.

nilsreichardt: Didn't find any issues for now.

app/lib/util/API.dart
```diff
@@ -66,18 +67,22 @@ class SharezoneGateway {
...
             authUser: authUser, firestore: references.firestore),
         fileSharing =
-            FileSharingGateway(user: authUser, references: references),
-        user = UserGateway(references, authUser);
+            FileSharingGateway(user: authUser, references: references);
+
+  Future<void> dispose() async {
+    await connectionsGateway.dispose();
+    await user.dispose();
+    blackboard.dispose();
```
Jonas-Sander, line 86: Why is there no await on the last call?

app/lib/main/run_app.dart
Jonas-Sander: Nit: Might it be more readable to do it as a pre-check via `if (user.typeOfUser == null)` instead of running it inside the `else` clause?

Also couldn't one do `userGateway?.dispose()`?

Jonas-Sander (approved):
See comments above, only nits though  - LGTM

app/lib/util/API.dart
Jonas-Sander: I would make the dispose method also async there so it matches the other calls

nilsreichardt: The `dispose` method of `BlackboardGateway` was sync but I saw that we can make it async. Therefore, I changed it to async so that we can await the call 👍

nilsreichardt:
> Nit: Might it be more readable to do it as a pre-check via `if (user.typeOfUser == null)` instead of running it inside the `else` clause?
> 
> Also couldn't one do `userGateway?.dispose()`?

I don't why but I can't answer to your comment, so I'm adding a new comment.

I would keep in the `else` clause and use `userGateway?.dispose()`

Issue #27, opened 2021-09-07
Add warning to alpha web app.


Jonas-Sander:
At https://alpha.web.sharezone.net we host an automatically deployed version from the main branch.     
We should make clear that we do not make any guarantee about anything in this version.   
There may be features not included by our privacy policy or tos and that there might be bugs that may lead to data loss. I guess for new features there might be even security risks (there might be ways to access data from other users if it haven't been secured properly) although I'm not sure how far we can go in this regard. I guess we still need to make some guarantees about security of ones data.

PR #216, opened 2022-05-11
Add banner when app is running an alpha version

nilsreichardt:
## Description
This PR contains a proposal to display an alpha version banner.

Users which are using an alpha version should be aware that is an alpha version. Therefore, a small banner at the top right-hand corner indicates this.

This banner also helps us when somebody sends a screenshot of an alpha version. We can instantly see that is an alpha version.

When this proposal got accepted, we can add the environment variable to the `deploy_web_app` workflow.

## Demo
image

## Related Tickets
Related to #27

The PR #220 should be merged first.


app/lib/widgets/alpha_version_banner.dart
```diff
@@ -0,0 +1,61 @@
...
+///
+/// This widget is similar and inspired by the [CheckedModeBanner] which display
+/// "DEBUG" at the top right hand corner when running a Flutter app in debug
+/// mode.
+class AlphaVersionBanner extends StatelessWidget {
+  /// Creates a const alpha version banner.
+  const AlphaVersionBanner({
+    Key key,
+    @required this.child,
+    this.isAlphaVersion = const bool.fromEnvironment('ALPHA'),
```
Jonas-Sander, line 29: What do you think about using `IS_ALPHA` instead of `ALPHA` for the environment variable?   

Or what about `--dart-define DEVELOPMENT_STAGE=ALPHA`?

Jonas-Sander: `ALPHA` is also okay for me, would just like to hear your opinion

app/lib/main/sharezone.dart
```diff
@@ -77,26 +78,28 @@ class _SharezoneState extends State<Sharezone> with WidgetsBindingObserver {
...
-                    return SharezoneApp(widget.blocDependencies,
-                        Sharezone.analytics, widget.beitrittsversuche);
-                  }
-                  return AuthApp(
-                    blocDependencies: widget.blocDependencies,
-                    analytics: Sharezone.analytics,
-                  );
-                },
+        child: AlphaVersionBanner(
+          child: Stack(
```
Jonas-Sander, line 82: (Not really important, just general thinking...)

In my head from looking at the widget tree I'm thinking in the first second that this banner will be always shown.
I'm often times asking myself if it's nicer to hide the implementation like you've done here or explicitly write:  
```dart
        child: AlphaVersionBanner(
          isAlphaVersion: bool.fromEnvironment('ALPHA'),
          child: Stack(
```
On the other hand we can't do this for every widget anyways and in many cases it would probably convolute the code if the logic is more complex than this case.

app/lib/widgets/alpha_version_banner.dart
Jonas-Sander, line 29:
```suggestion
    this.enabled = const bool.fromEnvironment('ALPHA'),
```
Or maybe `visible`?

WDYT? It would suggest that the UI is more "dumb".

app/lib/widgets/alpha_version_banner.dart
```diff
@@ -0,0 +1,61 @@
...
+    Widget result = child;
+    if (isAlphaVersion) {
+      result = Banner(
+        message: 'ALPHA',
+        textDirection: TextDirection.ltr,
+        location: BannerLocation.topEnd,
+        child: result,
+      );
+    }
+    return result;
```
Jonas-Sander, line 59:
```suggestion
    if (!isAlphaVersion) {
      return child;
    }

    return Banner(
      message: 'ALPHA',
      textDirection: TextDirection.ltr,
      location: BannerLocation.topEnd,
      child: child,
    );
```

Jonas-Sander (approved):
My comments are more or less nits.   
First golden test 🎉

In this case the advantage of the golden test is that it checks that the banner is in the right location, correct?

app/lib/widgets/alpha_version_banner.dart
Jonas-Sander: WDYT? Or why did you choose your pattern?

nilsreichardt: This looks much better 👍

nilsreichardt:
@Jonas-Sander What do you think about blue as a banner color?

| Dark mode | Light mode |
| - | - |
| image | image |
| image | image |

I like the blue 👍  What do you think? Or different color?

app/lib/widgets/alpha_version_banner.dart
nilsreichardt: Agree, much better to make the UI dumb 👍

app/lib/main/sharezone.dart
nilsreichardt: Changed it:
```dart
        child: AlphaVersionBanner(
          enabled: bool.fromEnvironment('ALPHA'),
          child: Stack(
```

app/lib/widgets/alpha_version_banner.dart
nilsreichardt: `DEVELOPMENT_STAGE=ALPHA` might have more option for future, like `BETA`, etc and has not the problem that `IS_BETA` and `IS_ALPHA` could be true.

app/lib/widgets/alpha_version_banner.dart
```diff
@@ -0,0 +1,60 @@
...
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+
+/// Displays a [Banner] saying "ALPHA" when running an alpha version at top left
+/// hand corner.
+///
+/// Displays nothing when [enabled] is false.
+///
+/// This widget is similar and inspired by the [CheckedModeBanner] which display
```
nilsreichardt, line 17:
```suggestion
/// This widget is similar and inspired by the [CheckedModeBanner] which displays
```

PR #138, opened 2022-03-23
Rework of the QR code scanner

nilsreichardt:
## Description

This PR removes the `barcode_scan` packages because it's deprecated and used old. For every build, we got this message:
```
The plugin `barcode_scan` uses a deprecated version of the Android embedding.
To avoid unexpected runtime failures, or future build failures, try to see if this plugin supports the Android V2 embedding. Otherwise, consider removing it since a future release of Flutter will remove these deprecated APIs.
If you are plugin author, take a look at the docs for migrating the plugin to the V2 embedding: https://flutter.dev/go/android-plugin-migration.
```

I created the internal package `qr_code_scanner`. It's just a wrapper around the [mobile_scanner](https://pub.dev/packages/mobile_scanner) package to add a custom UI and a few more features.

### Android

https://user-images.githubusercontent.com/24459435/193461163-0d800e57-4b11-4892-8194-df8423f627f1.mp4

### Web

https://user-images.githubusercontent.com/24459435/193463532-2c3fbc93-ed61-42e8-bac3-d9fb88cb7c24.mp4

### macOS

https://user-images.githubusercontent.com/24459435/193464889-18e39e79-6517-4b9b-bff8-e0bf12260080.mp4

## Todo
- [x] Add scan overlay with `CustomPainter` (UI like WhatsApp / Telegram scanning a QR code)
- [x] Add control for flashlight
- [x] Add subtitle
- [x] Add PR description
- [x] Add `README` for package
- [x] Manual test Android
- [x] Manual iOS
- [x] Manual macOS
- [x] Manual web
- [x] Add example
- [x]  Use scanner in `app`
- [x] Add license header
- [x] Handle phone landscape (landscape mode is very limited because `mobile_scanner` does not support it: https://github.com/juliansteenbakker/mobile_scanner/issues/107
- [ ] ~Maybe pass some kind of regex which is searched~ (I will this do this someday in the future)
- [x] Exception is thrown when leaving the page
- [x] App bar has wrong color in dark mode
- [x] Check landscape mode for iPad

## Related tickets
Closes #65

PR #220, opened 2022-05-11
Use for `test` job a macOS machine to be able to run golden tests

nilsreichardt:
#138 & #216 are adding golden tests. Golden tests can only run on one platform. We decided to pick macOS as the platform because primary using macOS. Therefore, should our CI use macOS machines. Otherwise, the Golden Tests are skipped.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,7 +59,7 @@ jobs:
       - run: addlicense -check -c "Sharezone UG (haftungsbeschränkt)" -f header_template.txt -ignore "**/GeneratedPluginRegistrant.swift" . 
 
   test:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -75,7 +75,7 @@ jobs:
         run: pub global activate --source path "$CI_CD_DART_SCRIPTS_PACKAGE_PATH"
 
       # So we can just use "sz COMMAND" instead of "dart ../path/to/script.dart ..."
-      - run: echo $(realpath ./bin) >> $GITHUB_PATH
+      - run: echo $(pwd)/bin >> $GITHUB_PATH
```
Jonas-Sander, line 78: Out of curiosity: Why did you change this? Did `realpath` lead to problems?

Jonas-Sander (approved):

PR #229, opened 2022-05-17
Add automatically labels to pull requests based on the changed file paths

nilsreichardt:
## Description
Finding pull requests with GitHub can be hard. Especially when our app has a bug and we try to find the pull request which caused the bug, might it be helpful to labels, like `platform: android` when we know it something with the Android build or changes to our CI/CD with the label `ci/cd`.

Therefore, I added the `labeler` Action to automatically add labels to our pull requests based on the changed file paths.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,3 +1,11 @@
+# Copyright (c) 2022 Sharezone UG (haftungsbeschränkt)
+# Licensed under the EUPL-1.2-or-later.
+#
+# You may obtain a copy of the Licence at:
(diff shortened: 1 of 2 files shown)


Jonas-Sander:
Much of the ci/cd code also lives inside our Sharezone CLI - we can never be sure though if a change really is only testing / ci/cd. Guess that's just a limitation we have to live with.

Jonas-Sander (approved):

nilsreichardt:
> we can never be sure though if a change really is only testing / ci/cd. Guess that's just a limitation we have to live with.

True, good to mention 👍

PR #231, opened 2022-05-17
Integrate iOS integration tests into CI pipeline

### LINKED ISSUE #177: Run integration tests for iOS in our CI

nilsreichardt:
Closes #177


nilsreichardt:
Status: Tests are not starting on the iOS simulator. I don't know why.

nilsreichardt:
@Jonas-Sander I found the issue. The `flutter test` times out because building the app takes long than the timeout (see comment in the code and https://github.com/flutter/flutter/issues/105913). Workaround is to use `flutter drive`. Ready to review 👍

.github/workflows/main.yml
```diff
@@ -114,6 +114,56 @@ jobs:
           # would be no failed golden test to upload.
           if-no-files-found: ignore   
   
+  ios-integration-test:
+    runs-on: macos-latest
+    timeout-minutes: 45
+    strategy:
+      matrix:
+        device:
+          - "iPhone 13"
```
nilsreichardt, line 123: Should we add more than the iPhone 13? Our project can run 5 macOS build simultaneously.

Jonas-Sander: Hm, idk. What do you think?

.github/workflows/main.yml
```diff
@@ -114,6 +114,56 @@ jobs:
...
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '11'
```
Jonas-Sander, line 130: For what is that required? Why the specific distribution?    
If its copied from an example somewhere then lets put the link in a comment above it. Otherwise in the future we might not know on the spot why its there and it might cause confusion.

Jonas-Sander (approved):
See the comment above. LGTM (provided the pipeline passes of course)

Jonas-Sander:
Wait does this also run on Draft PRs? It shouldn't I guess.

Jonas-Sander (changes requested):
See the two comments above

.github/workflows/main.yml
Jonas-Sander: If you want you can add more. I guess it might help to see if were breaking old iOS devices?

.github/workflows/main.yml
nilsreichardt: This can be removed 👍 It's only needed when using Flutter + Android. But for iOS we don't need this

.github/workflows/main.yml
nilsreichardt: I think one iPhone is for the beginning enough. We need also for the Android integration tests macOS machines, so I think we should not unnecessarily builds. If there is a reason to test on more OS version, then let's do this.

Jonas-Sander (approved):

Jonas-Sander:
You da boss 😎

PR #233, opened 2022-05-17
Use `--web-renderer canvaskit` flag instead of `--dart-define=FLUTTER_WEB_USE_SKIA=true` to deploy the web app via `sz deploy web-app`

nilsreichardt:
Regarding the official Flutter docs, you can set the renderer via the `--web-renderer` flag: https://docs.flutter.dev/development/platform-integration/web/renderers#command-line-options

### DIFF
--- a/tools/sz_repo_cli/lib/src/commands/src/deploy_web_app_command.dart
+++ b/tools/sz_repo_cli/lib/src/commands/src/deploy_web_app_command.dart
@@ -101,7 +101,8 @@ class DeployWebAppCommand extends Command {
           'build',
           'web',
           '--release',
-          '--dart-define=FLUTTER_WEB_USE_SKIA=true'
+          '--web-renderer',
+          'canvaskit',
         ],
         workingDirectory: _repo.sharezoneFlutterApp.location.path,
       );


tools/sz_repo_cli/lib/src/commands/src/deploy_web_app_command.dart
```diff
@@ -101,7 +101,8 @@ class DeployWebAppCommand extends Command {
           'build',
           'web',
           '--release',
-          '--dart-define=FLUTTER_WEB_USE_SKIA=true'
+          '--web-renderer',
+          'canvaskit',
```
Jonas-Sander, line 105: Wouldn't `auto` be better for mobile devices? Or is there sth that makes it unsuitable for us?

Jonas-Sander (changes requested):

tools/sz_repo_cli/lib/src/commands/src/deploy_web_app_command.dart
Jonas-Sander: We can use the canvas kit renderer for now, but back when we had the performance problems that was quite a while ago. We should test that again (might very well still be true of course).

Also what would be the overhead for "maintaining" it? I think it should more or less just work. Sure there is a slightly higher chance for bugs but if it works for us when we just scroll through once then I think the chance of new bugs occuring out of nowhere should be slim.

Jonas-Sander (approved):

tools/sz_repo_cli/lib/src/commands/src/deploy_web_app_command.dart
nilsreichardt: I agree that our goal should be using someday `auto` but with maintaining I mean we need to (manually) test the current features but also every upcoming feature.

Issue #166, opened 2022-05-06
Set up docs for Sharezone


nilsreichardt:
## Description
Currently, we use ZenDesk for our public docs and FAQ: https://sharezone.net/faq. Our goal is to move away from ZenDesk and use another Git based documentation tool, like [Riverpod](https://riverpod.dev/docs/getting_started), [FVM](https://fvm.app/docs/getting_started/overview) have.

The advantage is that it's free (instead of ZenDesk) and the community can easily participate ❤️ 

## ToDo
* [ ] Research for an open source documentation software (should also support multiple languages)
* [ ] Set up first proof of concept to make a decision
* [ ] Set up the software in this repository (+ hosting)
* [x] Set up CI/CD

nilsreichardt:
Basic docs are set up.

PR #235, opened 2022-05-17
Introduce the Android Alpha track 🎉

### LINKED ISSUE #215: Set up Alpha track for Android
## Description
We already have an alpha version for the web app: https://alpha.web.sharezone.net. We can set up exactly the same thing for Android with Firebase Distribution.

Similar to the alpha web app, we would **not** increase the build version or the app build code for every build because this would require a big overhead.

For every new release are the users notified that there is a new version available.

## Advantages
We can advertise this Alpha version to bring some users to this track. These users would use our app and might catch some bugs. This is helpful because we need a lot of time until we have a 100% test coverage (also with a lot of integration tests) to be really sure we have no bugs in our app.

nilsreichardt:
## Description
This PR adds an Alpha track for Android, like we already have for web with https://alpha.web.sharezone.net. On every commit on the `main`, we build a new Android version. Because Firebase Distribution doesn't care about versioning, we don't need a complicated system for increasing the version number.

As release note we are using the last commit message.

Users can join the Alpha track via this link: https://appdistribution.firebase.dev/i/9c4942a1c01a5496. Just follow the introductions by Google. But I think we should also add our own docs later (How to join the alpha program, how to leave, etc.). Would be good to do #166 first.

We should first merge the following PR before merging this PR:
* #241 

Closes #215


.github/workflows/alpha.yml
```diff
@@ -60,4 +48,75 @@ jobs:
...
+          --app 1:730263787697:android:f0a3b3b856fd1383 \
+          --groups "alpha" \
+          --release-notes "$LAST_COMMIT_MESSAGE"
+
+    - name: Remove credentials
+      if: always()
+      run: |
+        rm app/sharezone-prod-key.json
+        rm app/android/key.properties
+        rm app/android/app/key.jks
```
nilsreichardt, line 121: I don't know if this is really necessary but just the lower the security risk I add this step.

nilsreichardt:
@Jonas-Sander You can now review :)

.github/workflows/alpha.yml
```diff
@@ -60,4 +60,75 @@ jobs:
...
+        echo "keyPassword=$KEY_PASSWORD" >> key.properties
+        echo "keyAlias=$KEY_ALIAS" >> key.properties
+        echo "storeFile=key.jks" >> key.properties
+    
+    - name: Build Android
+      working-directory: app
+      run: | 
+        # We are publishing APKs instead of App Bundles to Firebase Distribution
+        # because they easier to install. App Bundles are installed via the
+        # PlayStore which resulted to problems in the fast.
```
Jonas-Sander, line 98:
```suggestion
        # PlayStore which resulted in problems in the past.
```

.github/workflows/alpha.yml
nilsreichardt: Yes

.github/workflows/alpha.yml
```diff
@@ -60,4 +60,75 @@ jobs:
...
+          --app 1:730263787697:android:f0a3b3b856fd1383 \
+          --groups "alpha" \
+          --release-notes "$LAST_COMMIT_MESSAGE"
+
+    - name: Remove credentials
+      if: always()
+      run: |
+        rm app/sharezone-prod-key.json
+        rm app/android/key.properties
+        rm app/android/app/key.jks
```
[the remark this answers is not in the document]
Jonas-Sander: Hm, okay. Can you create an issue for marking them as secrets? (I'm currently on mobile) Just so that we don't forget:)

Jonas-Sander (approved):
Please check again that nothing gets leaked in the logs and create the issue for marking the files as secrets (if the docs say it's necessary in our case).  

Otherwise LGTM! 🎉

.github/workflows/alpha.yml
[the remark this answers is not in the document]
nilsreichardt: I opened #245 but I think the in the section "Register all secrets used within workflows" they are talking about something different. I understand that if we create a **new** secret within a workflow (like generating a JWT), you can mark this as secret so that this secret with not shown in the logs. But all the files that we created like `sharzeone-prod-key.json`, `key.properties`, etc. are just already registered secrets (just written in a file). So they should already be masked in the logs.

My concern was a different one. I was more thinking about that the company GitHub is compromised and someone got access to the GitHub Actions containers and is able to read all the files. Or we say that we don't trust GitHub that they really delete all files after finishing a workflow. I know that are rare cases but my idea was to just decrease the risk where we have the impact to do it (especially when it's so easy like here we just execute 3 commands).

PR #230, opened 2022-05-17
Integrate Android integration tests into CI pipeline

nilsreichardt:
Closes #176

PR #236, opened 2022-05-17
Don't require `key.properties` anymore to build for Android

### LINKED ISSUE #234: `key.properties` is required even when building in debug mode
## Description
When you clone this repository and run `flutter run` for an Android device, you will get this error:
```
FAILURE: Build failed with an exception.

* Where:
Build file '/Users/nils/Desktop/Projects/sharezone/test/sharezone-app/app/android/app/build.gradle' line: 21

* What went wrong:
A problem occurred evaluating project ':app'.
> /Users/nils/Desktop/Projects/sharezone/test/sharezone-app/app/android/key.properties (No such file or directory)

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org
```

This is because the `app/build.gradle` tries to load the `key.properties` for Android code signing every time - no matter if you're building in release or debug mode.

https://github.com/SharezoneApp/sharezone-app/blob/8beeb01620eef3011668d3020e765f3857555fa3/app/android/app/build.gradle#L21

## Solution
We should only the load the `key.properties` when you really build in release mode.

nilsreichardt:
## Description
For now, it was required to have a `key.properties` file with our code signing credentials when you tried to build for Android (even when running in debug mode where no code signing was done).

We now check if a `key.properties` is present. If no, don't try to code sign the android app, which is the expected behavior because you need code signing only for publishing.

This PR unblocks #230 

Closes #234

### DIFF
--- a/app/android/app/build.gradle
+++ b/app/android/app/build.gradle
@@ -18,7 +18,8 @@ apply from: "$flutterRoot/packages/flutter_tools/gradle/flutter.gradle"
 
 def keystorePropertiesFile = rootProject.file("key.properties")
 def keystoreProperties = new Properties()
-if (keystorePropertiesFile.exists()) {
+def hasKeystorePropertiesFile = keystorePropertiesFile.exists()
+if (hasKeystorePropertiesFile) {
     keystoreProperties.load(new FileInputStream(keystorePropertiesFile))
 }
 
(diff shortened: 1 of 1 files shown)


Jonas-Sander (approved):
Very nice 👍

PR #253, opened 2022-05-21
Upload the results of failed golden tests

### LINKED ISSUE #252: Upload failed golden tests to GitHub Artifacts
## Description
Currently, when a test fails, the failed pictures are not uploaded to GitHub Artifacts. Therefore, it can be difficult to view the broken files, because they are not uploaded. Especially, when you are not using macOS, you need the uploaded results from CI (because golden tests are only running on macOS).

nilsreichardt:
## Description
Uploads the results of a failed golden tests to GitHub, like: https://github.com/SharezoneApp/sharezone-app/actions/runs/2363668983

Closes #252

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -95,6 +95,20 @@ jobs:
       
       - name: Run tests via "sz test"
         run: sz test -v
+      
+      - name: Upload failed golden tests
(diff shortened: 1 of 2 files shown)


Jonas-Sander (approved):
Wow very nice

PR #257, opened 2022-06-05
Add Codemagic App Previews (Android & iOS) for Pull Requests

nilsreichardt:
## Description

We are already using the [Deploy to Firebase](https://github.com/marketplace/actions/deploy-to-firebase-hosting) action to deploy our web to a temporary url, when opening a new pull request.

[codemagic_app_preview](https://pub.dev/packages/codemagic_app_preview) does the same for iOS and Android. Just scan the QR with your phone, install the app and review the changes 👍 

Currently, I'm not able to build a iOS and macOS with one Codemagic workflow. Therefore, we are not building the macOS app at the moment. When I fixed this, I will make a new PR for adding support for macOS 👍 

Google Sign In & Dynamic Links are not working at the moment for the Android version, because the dev Firebase project does not contain the SHA1 & SHA256 values of the code signing key (see #289).


nilsreichardt:
⬇️ Generated builds by [Codemagic](https://codemagic.io/app/629c82ea463af7ff553fc7a5/build/6307b4fcee1f41ccd749d5d5) for commit 43de63dff3735f9e5d8e44074859c48f9ccf0430 ⬇️

| Android |
|:-:|
| image <br /> [Download link](https://api.codemagic.io/artifacts/dc3b5b73-051b-4d48-9343-d7f59c4e303a/b1eb4e37-124a-46b4-b67a-0a70d4812fd0/app-dev-debug.apk) |

Jonas-Sander:
Hey, this is just a quick thought written down, not a review yet.  

Since the `codemagic.yaml` code uses some secretes for signing that are very important - might there be some way for an external "attacker" to read them by forking our repository and proposing a PR with changes inside `codemagic.yaml` which will send the secrets to the attacker?  

So under which conditions does the `codemagic.yaml` get executed? If an external user were to propose changes do we need to manually approve running the `codemagic.yaml`? (couldn't the user still do the same pwning-changes after manual approval?)

I read this article today which talks about something similar: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

codemagic.yaml
```diff
@@ -0,0 +1,79 @@
...
+    max_build_duration: 75
+    environment:
+      groups:
+        - github
+        - ios-publishing
+        - android_publishing
+        - certificate_credentials
+      vars:
+        BUNDLE_ID: de.codingbrain.sharezone.app
+      flutter: v2.10.5
```
Jonas-Sander, line 39: Can this somehow be read from the fvm config or `pubspec,yaml`, right?

codemagic.yaml
```diff
@@ -0,0 +1,79 @@
...
+    publishing:
+      scripts:
+        - name: Add Dart SDK to PATH
+          script: |
+            echo PATH="$PATH":"$FLUTTER_ROOT/.pub-cache/bin" >> $CM_ENV
+            echo PATH="$PATH":"$FLUTTER_ROOT/bin" >> $CM_ENV
+        - name: Post App Preview
+          script: |
+            dart pub global activate codemagic_app_preview
+            app_preview post --gh_token $GITHUB_PAT --message "_Note_: Only Sharezone team members are able to install the iOS app."
```
Jonas-Sander, line 79: > codemagic_app_preview

They grow up so fast, now they even got their own published package we're using 😭😂

nilsreichardt:
Good point about the security. I tried out to print a secret and it was easily possible. I asked the Codemagic support what you can to do to protect your secrets.

codemagic.yaml
nilsreichardt: Technically, this should be possible, but it would take an overhead of ~1-2 minutes.

What Codemagic does:
1. Download specified Flutter version
2. Clone the repository

Therefore, we would need to this:

1. Codemagic downloads the specified (or the default one) Flutter version
2. Codemagic clones the repository
3. We read out the Flutter version specified in the `fvm_config.json`
4. Download the Flutter version
5. Update the environment variables

Jonas-Sander: Hm that kinda sucks. Do we have a document "update the flutter version" or sth where we can for now at least have the places written down where we would need to update the version string?

Jonas-Sander:
The pipeline is failing though, so only merge if everything works ofc;)

nilsreichardt:
> From our conversation: This code is not a security risk if an external contributor changes the code inside the codemagic.yaml to send secrets to themselves because codemagic will use the codemagic.yaml from the default branch (main) (but only if the PR comes from an forked repositoy) - correct?

Correct 👍

nilsreichardt:
> The pipeline is failing though, so only merge if everything works ofc;)

I changed nothing and iOS code signing is not working anymore correctly 😭

codemagic.yaml
nilsreichardt: Using FVM seems to be a good option 👍 The overhead is roundabout 1 min

nilsreichardt:
@Jonas-Sander I added a few things. Can you re-review? You can simply use the feature "show changes since last review"

codemagic.yaml
```diff
@@ -0,0 +1,110 @@
...
+      vars:
+        BUNDLE_ID: de.codingbrain.sharezone.app.dev
+      # We won't use this Flutter version because we are going to download FVM
+      # later and use the specified Flutter version in our repository.
+      #
+      # As predefined Flutter version we are using here the newest stable
+      # version because this version is preinstalled on Codemagic VMs.
+      flutter: stable
+    when:
+      # To save Codemagic usage we only run the tests for non draft PRs
```
Jonas-Sander, line 54:
```suggestion
      # To save Codemagic usage we only build non draft PRs
```

Jonas-Sander (approved):
LGTM 🙌

PR #266, opened 2022-06-27
Use `themeMode` in SharezoneMaterialApp.

Jonas-Sander:
From `themeMode` docs:
> Determines which theme will be used by the application if both [theme] and [darkTheme] are provided.
> 
> If set to [ThemeMode.system], the choice of which theme to use will be based on the user's system preferences. If the [MediaQuery.platformBrightnessOf] is [Brightness.light], [theme] will be used. If it is [Brightness.dark], [darkTheme] will be used (unless it is null, in which case [theme] will be used.
> 
> If set to [ThemeMode.light] the [theme] will always be used, regardless of the user's system preference.
> 
> If set to [ThemeMode.dark] the [darkTheme] will be used regardless of the user's system preference. If [darkTheme] is null then it will fallback to using [theme].
> 
> The default value is [ThemeMode.system].

### DIFF
--- a/app/lib/main/sharezone_material_app.dart
+++ b/app/lib/main/sharezone_material_app.dart
@@ -82,7 +82,6 @@ ThemeMode _getThemeMode(ThemeBrightness themeBrightness) {
       return ThemeMode.light;
     case ThemeBrightness.system:
       return ThemeMode.system;
-    default:
   }
   throw UnimplementedError();
 }
--- a/app/lib/main/sharezone_material_app.dart
+++ b/app/lib/main/sharezone_material_app.dart
@@ -49,12 +49,9 @@ class SharezoneMaterialApp extends StatelessWidget {
         debugShowCheckedModeBanner: false,
         title: PlatformCheck.isWeb ? "Sharezone Web-App" : "Sharezone",
         color: primaryColor,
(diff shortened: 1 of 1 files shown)


app/lib/main/sharezone_material_app.dart
```diff
@@ -76,3 +73,16 @@ class SharezoneMaterialApp extends StatelessWidget {
...
+
+ThemeMode _getThemeMode(ThemeBrightness themeBrightness) {
+  switch (themeBrightness) {
+    case ThemeBrightness.dark:
+      return ThemeMode.dark;
+    case ThemeBrightness.light:
+      return ThemeMode.light;
+    case ThemeBrightness.system:
+      return ThemeMode.system;
+    default:
```
nilsreichardt, line 85: Would it make more sense to don't have a `default` value? Therefore, the analyzer would show a warning when add a new value for `ThemeBrithness`

nilsreichardt (approved):
LGTM

app/lib/main/sharezone_material_app.dart
Jonas-Sander: Good catch! Yeah of course, was maybe left over from the standard template.

Jonas-Sander, line 85:
```suggestion
```

PR #267, opened 2022-07-01
Remove unused "lessonChanges" from events.

Jonas-Sander:


Jonas-Sander:
@nilsreichardt Do you know if this has actually ever been used?

nilsreichardt (approved):
LGTM

No, we never used it.

I also tested that this does not break old clients 👍

PR #292, opened 2022-09-27
Update main.yml

Test-Banane:


Test-Banane:
Just tested to print secrets (Nils)

PR #297, opened 2022-09-27
Fix "label-pr" workflow for PR from forks

### LINKED ISSUE #294: Workflow "label-pr" is not working when a PR is opened from a fork
Needs to be changed into `pull_request_target`

nilsreichardt:
Tested it manually (by using `pull_request`): https://github.com/SharezoneApp/sharezone-app/actions/runs/3136587111/jobs/5093688752

The workflow is not running for this PR because it's not on our `main` branch. It will be used when this PR has been merged.

Closes #294


Jonas-Sander (changes requested):

.github/workflows/pull_request_label.yml
```diff
@@ -0,0 +1,45 @@
...
+      - opened
+      - synchronize
+      - reopened
+      - closed
+
+# Declare default permissions as read only.
+#
+# Having workflows without default permissions is considered a bad security
+# practice and it is causing alerts from our scanning tools.
+permissions: read-all
```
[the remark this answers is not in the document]
nilsreichardt: Because the repository is public, I don't think you can not read with the `GITHUB_TOKEN` things that are not already public. But I changed it to `permissions: {}` which disables all permission and it still works 👍

.github/workflows/pull_request_label.yml
```diff
@@ -0,0 +1,45 @@
...
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - closed
+
+# Declare default permissions as none.
+#
+# Having workflows without default permissions is considered a bad security
+# practice and it is causing alerts from our scanning tools.
```
Jonas-Sander, line 26:
```suggestion
# Having workflows with default permissions is considered a bad security
# practice and it is causing alerts from our scanning tools.
```

Shouldn't it be this?

Jonas-Sander (approved):
LGTM, See comment above though

.github/workflows/pull_request_label.yml
nilsreichardt: It means the default permission in workflow. In this case

```yaml
permissions: {}
```

PR #300, opened 2022-09-28
Set permissions for workflows to follow a good security practice.

nilsreichardt:
Having workflows without setting the permission for the workflow is considered a bad security practice and it is causing alerts from our scanning tools.

### DIFF
--- a/.github/workflows/alpha.yml
+++ b/.github/workflows/alpha.yml
@@ -29,12 +29,11 @@ on:
 env:
   CI_CD_DART_SCRIPTS_PACKAGE_PATH: "tools/sz_repo_cli/"
 
-# Declare default permissions as none.
+# Set permissions to none.
 #
-# Having workflows without setting the permission for the workflow is considered
-# a bad security practice and it is causing alerts from our scanning tools.
+# Using the broad default permissions is considered a bad security practice 
+# and would cause alerts from our scanning tools.
 permissions: {}
-
 jobs:
   deploy-alpha-web-app:
     runs-on: ubuntu-latest
--- a/.github/workflows/alpha.yml
+++ b/.github/workflows/alpha.yml
@@ -29,6 +29,12 @@ on:
 env:
   CI_CD_DART_SCRIPTS_PACKAGE_PATH: "tools/sz_repo_cli/"
 
+# Declare default permissions as none.
+#
+# Having workflows without setting the permission for the workflow is considered
+# a bad security practice and it is causing alerts from our scanning tools.
+permissions: {}
+
 jobs:
   deploy-alpha-web-app:
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -162,7 +162,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write # for FirebaseExtended/action-hosting-deploy to comment on PRs
-      checks: write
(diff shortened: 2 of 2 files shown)


.github/workflows/alpha.yml
```diff
@@ -29,6 +29,12 @@ on:
 env:
   CI_CD_DART_SCRIPTS_PACKAGE_PATH: "tools/sz_repo_cli/"
 
+# Declare default permissions as none.
+#
+# Having workflows without setting the permission for the workflow is considered
+# a bad security practice and it is causing alerts from our scanning tools.
+permissions: {}
+
```
Jonas-Sander, line 37:
```suggestion
# Set permissions to none.
#
# Using the broad default permissions is considered a bad security practice 
# and would cause alerts from our scanning tools.
permissions: {}
```

.github/workflows/main.yml
```diff
@@ -21,6 +21,12 @@ on:
 env:
   CI_CD_DART_SCRIPTS_PACKAGE_PATH: "tools/sz_repo_cli/"
 
+# Declare default permissions as none.
+#
+# Having workflows without setting the permission for the workflow is considered
+# a bad security practice and it is causing alerts from our scanning tools.
+permissions: {}
```
Jonas-Sander, line 28:
```suggestion
# Set permissions to none.
#
# Using the broad default permissions is considered a bad security practice 
# and would cause alerts from our scanning tools.
permissions: {}
```

.github/workflows/main.yml
```diff
@@ -154,6 +160,9 @@ jobs:
   # 4. Adjust website restrictions for Firebase Key "Sharezone Web Key".
   web-preview:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write # for FirebaseExtended/action-hosting-deploy to comment on PRs
+      checks: write # for FirebaseExtended/action-hosting-deploy
```
Jonas-Sander, line 165: Couldn't find what the `checks` permissions is exactly for. Can you add it to the comment next to it?

Jonas-Sander (changes requested):

.github/workflows/main.yml
nilsreichardt: The thing is that I also don't know why it's needed but without it's not posting the comment. It also needs to be `write`. `read` is not working.

nilsreichardt: Added the comment

> for FirebaseExtended/action-hosting-deploy to comment on PRs (without write permissions for checks the action doesn't post a comment to the PR)

.github/workflows/main.yml
```diff
@@ -154,6 +160,9 @@ jobs:
   # 4. Adjust website restrictions for Firebase Key "Sharezone Web Key".
   web-preview:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write # for FirebaseExtended/action-hosting-deploy to comment on PRs
+      checks: write # for FirebaseExtended/action-hosting-deploy to comment on PRs (without write permissions for checks the action doesn't post a comment to the PR)
```
Jonas-Sander, line 165:
```suggestion
      checks: write # for FirebaseExtended/action-hosting-deploy to comment on PRs (without write permissions for checks the action doesn't post a comment to the PR, we don't know why)
```

I think it's better to make it clear we don't know why so that in the future one might be more inclined to investigate than see the `checks` permission as a given that can't be changed (for all we know right now it might be a bug)

Jonas-Sander (approved):
LGTM

PR #319, opened 2022-10-02
Assign label "dependencies" automatically

nilsreichardt:


Jonas-Sander:
But we're importing all our `/lib` packages there as well?

nilsreichardt:
> But we're importing all our `/lib` packages there as well?

It's intended when a dependency changes in `lib` that it should get the label. Not sure if "**external** dependency" is maybe a bit misleading and just "dependencies" is better, because a PR should also get the label when `app` imports an internal package

Jonas-Sander:
Okay then I understood it wrong, then change it back to just dependencies if you want, sry

Jonas-Sander (approved):

nilsreichardt:
I also forgot about this case in the PR 😅

PR #376, opened 2023-02-24
Initialize Firebase via Dart

### LINKED ISSUE #221: Initialize Firebase via Dart to run the newest Firebase packages in web
## Description
Currently, we add the Firebase credentials via JavaScript:
https://github.com/SharezoneApp/sharezone-app/blob/324e90a1c8e698b43c2518f8ce1d20f4db8f8405/app/web/index.html#L67

This is not needed anymore with newer Firebase packages because they use the credentials from the Dart initialization.

See https://firebase.google.com/docs/flutter/setup how to set up Firebase via Dart.

### LINKED ISSUE #375: Fix `sz deploy web-app` command.
Since #369 using the deploy command for the web app is broken since the Google credentials are not imported via the `index.html` anymore, but via Dart code. This means that all web apps uploaded via this method will be using the debug/dev project.

nilsreichardt:
## Description

With this PR we initialize Firebase also via Dart. This change requires setting the `main` file when running the Flutter app:

```sh
flutter run \
  --flavor dev \
  --target lib/main_dev.dart
```

This is because we need a way to know which flavor is used inside the Dart code to terminate which Firebase project we should initialize. An alternative would be to use `--dart-define`. However, I like the `--target` solution more because there is no default option and you have to explicitly set a flavor. With `--dart-define`, we would either need to set a default or throw an exception at runtime.

Our file structure looks like this:
```
lib/
  ...
  main_common.dart
  main_dev.dart
  main_prod.dart
```

`main_common.dart` is just the old `main.dart` but with a required `flavor` parameter.

We are going to keep the native Firebase initialization for Android on iOS because we need it for Crashlytics and from my experience is the initialize Firebase way for Android and iOS not that stable. It's no problem to have both solutions.

## Related Tickets

Closes #375 
Closes #221


.vscode/launch.json
```diff
@@ -4,32 +4,15 @@
...
-        },
-        {
-            "name": "[Release] main",
-            "type": "dart",
-            "request": "launch",
-            "args": [
-                "--release",
-            ],
-            "program": "app/lib/main.dart"
-        },
```
nilsreichardt, line 23: We don't need them anymore, because we can now specify the flavors for all platforms via Dart.

Sharezone-Admin:
⬇️ Generated builds by [Codemagic](https://codemagic.io/app/629c82ea463af7ff553fc7a5/build/63f942f6d8e0c9b2af45eeda) for commit 7ced69bbb55527e0f59686a3d01e626980a09bd8 ⬇️

|
|
|

Jonas-Sander (approved):
Very nice!

Jonas-Sander:
> We are going to keep the native Firebase initialization for Android on iOS because we need it for Crashlytics and from my experience is the initialize Firebase way for Android and iOS not that stable. It's no problem to have both solutions.

From what I can see you use the Dart initialization for all platforms now, is the description out-of-date or am I just confused right now?

nilsreichardt:
We are using the Dart initialization for all platforms, yes. But we still have some native Firebase setup for Android and iOS, like `/app/android/app/src/prod/google-services.json`. So for Android and iOS we are using the old and new way. We are using both ways because it's more stable and for some services (like Crashlytics) you still need the native setup. In the Dart code we don't need to exclude Android & iOS.

PR #404, opened 2023-02-28
Add Fallback App CI

### LINKED ISSUE #403: App workflows waiting for every when they are not triggered
image

With #400 we only trigger the app workflows when there was an app change. However, when there was not an app change, the required workflows are not triggered and it's impossible to merge the PR.

nilsreichardt:
image

A workaround for GitHub Actions jobs that required and based on a path.

Currently, it's not possible to set a GitHub Action to required which is based on path changes. The reason for this that the action is not executed when the action does not match the paths. Therefore, merging will be blocked.

A workaround is to have a job which the same name. 

See: https://github.com/github/docs/commit/4364076e0fb56c2579ae90cd048939eaa2c18954

Closes #403

### DIFF
--- a/.github/workflows/app_ci.yml
+++ b/.github/workflows/app_ci.yml
@@ -26,6 +26,9 @@ on:
       # ready for review.
       - ready_for_review
     paths:
+      # When you add a new path, make sure to also add the path to the
+      # "paths-ignore" section in the "app_fallback_ci.yml" file.
+      # 
       # We only build and deploy a new version, when a user relevant files
       # changed.
       - "app/**"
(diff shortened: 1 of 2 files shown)


nilsreichardt:
@Jonas-Sander Could you also review this PR because this blocks #407

Jonas-Sander (approved):

Jonas-Sander:
btw we need to execute the jobs when changing the sharezone cli folder aswell (since some call the functions in there)

nilsreichardt:
> btw we need to execute the jobs when changing the sharezone cli folder aswell (since some call the functions in there)

Good point, I'll make a new PR for this 👍

PR #407, opened 2023-03-02
Add docs for our alpha program

nilsreichardt:
Adds a first doc page about our alpha program. It's just a first version. When I have more time, I would like to add more information, add video tutorials on how to join the alpha program, etc.

Docs are in German because our has only German as language. 

image

Preview the docs: https://docs.sharezone.net/~407/more/alpha-program


docs.json
```diff
@@ -9,6 +9,12 @@
     [
       "Another Page",
       "/another-page"
+    ],
+    [
+      "Weiteres",
+      [
+        ["Alpha-Programm", "/more/alpha-program"]
+      ]
```
nilsreichardt, line 17: My idea is to have for every feature (like homework, timetable, events, etc.) a own section and a section "more" where we have non feature things (like the alpha program). Later, we might going to split the "more" section if it becomes to big.

Jonas-Sander: Hm, okay, the only thing is that the URL might not be stable then if we split off the article later to some other section.  
Okay with me though, since we don't have an existing, better way to categorize the articles.

docs/more/alpha-program.mdx
```diff
@@ -0,0 +1,70 @@
+# Alpha-Programm
+
+Wir freuen uns sehr, dass du dich für das Alpha-Programm von Sharezone
+interessierst! Mit Alpha-Version unserer App bieten wir dir die Möglichkeit,
```
Jonas-Sander, line 4:
```suggestion
interessierst! Mit der Alpha-Version unserer App bieten wir dir die Möglichkeit,
```

docs/more/alpha-program.mdx
```diff
@@ -0,0 +1,70 @@
...
+funktioniert.
+
+## Wie melde ich mich für das Alpha-Programm an?
+
+### Android
+
+Um die Alpha-Version unserer App auf Android-Geräten zu testen, musst du dich
+zunächst mit [diesem Link](https://sharezone.net/android-alpha) anmelden.
+
+Auf dieser Seite kannst du mich mit einer E-Mail-Adresse für das Alpha-Programm
```
Jonas-Sander, line 35:
```suggestion
Auf dieser Seite kannst du dich mit einer E-Mail-Adresse für das Alpha-Programm
```

Jonas-Sander (approved):
Looks good, very nice! :)

docs.json
nilsreichardt: Yes, I also don't know to fix that really. I also wanted to create a shorter link for some things, like our alpha program: https://sharezone.net/alpha which would redirect to the docs. At least we could update there the link. However, I think our traffic for the docs will not that much high, so I think it's okay.

PR #408, opened 2023-03-03
Add CI for Sharezone CLI

nilsreichardt:
This PR adds a CI for our CLI because with #400 the CLI was excluded from our CI. 

I have not added the path for our CLI to our `app_ci.yml` because otherwise we would also trigger the integration test and the CI for our app, which is not needed when only modifying our CLI. This makes our CI faster and we don't need to use a `macos` machine for our `test` job here.

A side effect is currently that we also run the analyze and test step for our CLI in our app CI. However, I think that is not a big problem (in the future we could add a `exclude-package` argument to the `sz analyze` and `sz test` command to prevent this).


Jonas-Sander:
Hm, I don't know, I dislike this solution as we're not using the `sz` commands to test and analyze. So instead of using `sz test` `dart test` is used.    

I would like to always use the CLI test code since if we make a change there it is used everywhere. For example lets imagine we would want to use the `--test-randomize-ordering-seed` flag for running all our Dart/Flutter test. In this case if we only use the CLI code/commands we only need to change it in one place. With your approach we would also need to remember to change it here as well.

Couldn't we for now just run it together with the `/lib` libraries and maybe add an `--only tools/sz_repo_cli` flag or sth similar`in the future?

nilsreichardt:
Ah, yes makes sense to use `sz test` and in the future with `--only tools/sz_repo_cli`. I changed to use `sz test` and `sz analyse`.

Jonas-Sander:
But wouldn't it now run two times if you change something in the `tools` and the `lib` folder?

Sharezone-Admin:
⬇️ Generated builds by [Codemagic](https://codemagic.io/app/629c82ea463af7ff553fc7a5/build/640762959e31ee0341f1592b) for commit ec0b2b213f03f951e55073eb8524171cf8ca494e ⬇️

| iOS | Android |
|:-:|:-:|
| image <br /> [Download link](https://api.codemagic.io/artifacts/f03dbb5f-e5d3-48b7-b09f-f343700ac782/10e53812-58d1-486b-9a2e-b4f5fb12d5f9/sharezone.ipa) | image <br /> [Download link](https://api.codemagic.io/artifacts/ab5abbb1-7b72-42be-a228-d0edf26ced9f/ba2f6ebe-ccad-4d80-b315-a3f524666fb6/app-dev-release.apk) |

nilsreichardt:
> So why not just add the tools folder to the app_ci.yml?

Because this would run the following workflows even when we only modify the CLI code:

* `web-preview`
* `android-integration-test`
* `ios-integration-test`
* (in the future) `mac-build-test`

nilsreichardt:
> But wouldn't it now run two times if you change something in the tools and the lib folder?

Yes, this is the trade off of:

> Couldn't we for now just run it together with the /lib libraries and maybe add an --only tools/sz_repo_cli flag or sth similar`in the future?

Jonas-Sander (approved):

Jonas-Sander:
What do you think of the `--only tools/sz_repo_cli` flag solution for the future?

PR #418, opened 2023-03-06
Migrate `remote_configuration` to null safety

### LINKED ISSUE #188: Migrate the `remote_configuration` package to null safety
Blocked by:
* #186

nilsreichardt:
Closes #188


lib/remote_configuration/lib/src/implementation/firebase_remote_configuration.dart
```diff
@@ -12,30 +12,37 @@ import 'package:sharezone_utils/platform.dart';
...
 
+  FirebaseRemoteConfig? _remoteConfig;
+  Map<String, dynamic>? _defaultValues;
+
   @override
   String getString(String key) {
-    return _remoteConfig.getString(key);
+    if (_remoteConfig == null) {
+      throw RemoteConfigurationNotInitializedException();
+    }
```
nilsreichardt, line 24: What should we do when `_remoteConfig` or `_defaultValues` are not initialized?

1. Throw a `RemoteConfigurationNotInitializedException`?
2. Return null?
3. Make `_remoteConfig` and `_defaultValues` `late` and Dart would throw an exception that `_remoteConfig` has not been initialized

Jonas-Sander (changes requested):
See comment above

lib/remote_configuration/lib/src/implementation/firebase_remote_configuration.dart
nilsreichardt: > _defaultValues should never be null, if none are passed then this should just be treated as passing an empty map in my opinion.

`_defaultValues` are only used in our `StubRemoteConfiguration`. Treating the `defaultValues` like an empty map would required that we are returning default values for keys that does not exit because otherwise we would get a runtime exception that `null` is not `String` or `bool`:

```dart
class StubRemoteConfiguration extends RemoteConfiguration {
  Map<String, dynamic> _defaultValues = {};

  String getString(String key) {
    return _defaultValues[key] ?? '';
  }

  @override
  bool getBool(String key) {
    return _defaultValues[key] ?? false;
  }

  @override
  Future<void> initialize(Map<String, dynamic> defaultValues) async {
    _defaultValues = defaultValues;
  }
}
```

What do you think about this?

nilsreichardt: > _defaultValues should never be null, if none are passed then this should just be treated as passing an empty map in my opinion.

`_defaultValues` are only used in our `StubRemoteConfiguration`. Treating the `defaultValues` like an empty map would required that we are returning default values for keys that does not exit because otherwise we would get a runtime exception that `null` is not `String` or `bool`:

```dart
class StubRemoteConfiguration extends RemoteConfiguration {
  Map<String, dynamic> _defaultValues = {};

  String getString(String key) {
    return _defaultValues[key] ?? '';
  }

  @override
  bool getBool(String key) {
    return _defaultValues[key] ?? false;
  }

  @override
  Future<void> initialize(Map<String, dynamic> defaultValues) async {
    _defaultValues = defaultValues;
  }
}
```

What do you think about this?

nilsreichardt: (but I have also to say that `StubRemoteConfiguration` isn't used anymore and could also be removed)

Jonas-Sander: Why would the classes return a `String` instead of a `String?` That doesn't make sense at all.    
The interface is wrong.

Jonas-Sander: wait what, but what does it return if no value is found for the given key?

nilsreichardt: Firebase returns default values (like `""` for Strings) which can find here: https://firebase.google.com/docs/reference/android/com/google/firebase/remoteconfig/FirebaseRemoteConfig#constants_1

Jonas-Sander: wwwtttttffffffffff

Jonas-Sander: ok lets just follow this behavior (for now) ._.    
I don't like it though

Jonas-Sander (approved):

Issue #37, opened 2021-09-11
Use the `sharezone_lints` for all packages.


Jonas-Sander:
E.g. the `lint` package.

Jonas-Sander:
> It should be fixed with #37

Should be fixed with what? #37 is this issue.
Also some packages had different analysis_options from what I remember so we would have to fix some lints to achieve this I think

nilsreichardt:
Oh, I wanted to use #316. However, we can still keep this issue open and maybe change the title to use something like "Use `sharezone_lints` in all package"

PR #423, opened 2023-03-07
Add `sharezone_lints` to `notifications` package

nilsreichardt:
This PR fixes the following warnings:

```
Analyzing notifications...                                              

   info • Use the generic function type syntax in 'typedef's • lib/src/action_registration.dart:24:24 • prefer_generic_function_type_aliases
   info • The imported package 'meta' isn't a dependency of the importing package • lib/src/instrumentation.dart:9:8 •
          depend_on_referenced_packages
   info • The local variable '_instrumentation' starts with an underscore • lib/src/notification_parser.dart:21:50 •
          no_leading_underscores_for_local_identifiers
   info • The imported package 'collection' isn't a dependency of the importing package • lib/src/push_notification.dart:9:8 •
          depend_on_referenced_packages
   info • Unnecessary 'this.' qualifier • lib/src/push_notification.dart:51:8 • unnecessary_this
   info • The local variable '_other' starts with an underscore • lib/src/push_notification_action_handler.dart:216:43 •
          no_leading_underscores_for_local_identifiers
   info • The local variable '_duplicates' starts with an underscore • lib/src/push_notification_action_handler.dart:223:29 •
          no_leading_underscores_for_local_identifiers
   info • Don't invoke 'print' in production code • test/docs_example_test.dart:24:9 • avoid_print
   info • Use 'const' with the constructor to improve performance • test/docs_example_test.dart:36:7 • prefer_const_constructors

9 issues found. (ran in 1.4s)
```

Part of #37

### DIFF
--- a/lib/notifications/test/docs_example_test.dart
+++ b/lib/notifications/test/docs_example_test.dart
@@ -6,6 +6,8 @@
 //
 // SPDX-License-Identifier: EUPL-1.2
 
+import 'dart:developer';
+
 import 'package:test/test.dart';
 import 'package:notifications/notifications.dart';
 
@@ -21,7 +23,7 @@ void main() {
         );
       },
       executeActionRequest: (request) {
-        print(request.stringToPrint);
+        log(request.stringToPrint);
       },
     );
 
@@ -33,7 +35,7 @@ void main() {
     );
 
     handler.handlePushNotification(
-      PushNotification(
+      const PushNotification(
         actionType: 'print-secret-string',
         actionData: {'secret-message': 'SHAREZONE4EVER'},
         title: 'Title',
--- a/lib/notifications/test/docs_example_test.dart
+++ b/lib/notifications/test/docs_example_test.dart
@@ -6,10 +6,13 @@
 //
 // SPDX-License-Identifier: EUPL-1.2
 
-import 'dart:developer';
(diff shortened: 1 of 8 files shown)


lib/notifications/test/docs_example_test.dart
```diff
@@ -21,7 +23,7 @@ void main() {
         );
       },
       executeActionRequest: (request) {
-        print(request.stringToPrint);
+        log(request.stringToPrint);
       },
     );
```
Jonas-Sander, line 28: Well the thing is this is the code example for the documentation of the `PushNotificationActionHandler`.  
So I guess the example in the doc also needs to be updated, altough I think it would be more confusing to use log instead of print in the example.  
So maybe just add `ignore_for_file: lint_name` to the top of this file?

Jonas-Sander (approved):
See above

lib/notifications/test/docs_example_test.dart
nilsreichardt: Yes, makes sense. I added it in https://github.com/SharezoneApp/sharezone-app/pull/423/commits/1f6b5c772f55780f74de976ab65c35baaafbb9d9

Jonas-Sander (approved):

PR #424, opened 2023-03-07
Migrate `common_domain_model` package to null safety

### LINKED ISSUE #203: Migrate the `common_domain_models` package to null safety
Blocked by:
* #181

nilsreichardt:
Closes #203


lib/common_domain_models/lib/src/ids/abgabenziel_id.dart
```diff
@@ -71,6 +71,7 @@ class AbgabezielId extends Id {
         _id = HomeworkId(contentIdString);
         break;
       default:
+        _id = Id(contentIdString);
     }
```
Jonas-Sander, line 76: Does this make sense? I guess you just added it so `_id` is not null, but idk if that is correct behavior.   
For now we only have implemented it for homeworks anyway. Should we not rather throw an error?

Jonas-Sander (changes requested):

lib/common_domain_models/lib/src/ids/abgabenziel_id.dart
nilsreichardt: Yes, for now, we have it only for homework. I'm going to change to throw an error 👍

Jonas-Sander (approved):

nilsreichardt:
In the future we always should have an `unknown` value in our enums (for enums that are coming from the database) because when we add types and throwing errors for unknown types, will this break on old clients

This is how it looks right now:
```dart
enum AbgabenzielTyp { hausaufgabe }
```

This is how it should look like:
```dart
enum AbgabenzielTyp { hausaufgabe, unknown }
```

And then we display in the UI something "Can display XYZ, maybe your app version is too old" when we use `unknown`.

PR #436, opened 2023-03-08
Remove unused code in `build_context` package

nilsreichardt:


Jonas-Sander:
I mean its in and of itself fine with me to remove the unused code, but does it really make sense to remove all the extension attributes just because we don't use them _right now_? It just feels weird, I mean we would have to re-add it everytime we want to use another attribute (and probably I would just not think of it and do it the "long way" without extension). For me this is kinda different to usual unused code. WDYT?

nilsreichardt:
I'm unsure if these extensions are really helpful. I would also like to remove `theme`, `textScaleFactor`, `hideCurrentSnackBar `, `scaffoldBackgroundColor`, `mediaQueryViewPadding`, `primaryColor` but they are used in code at the moment. The only helpful extension attributes which make the code more readable are `isDesktopModus` and `isDarkModeEnabled`. All other extension methods and attributes are just saving the `X.of(context)`. WDYT?

Jonas-Sander:
> All other extension methods and attributes are just saving the X.of(context). 

I mean thats why the extensions were created in the first place, no? :D  

> I'm unsure if these extensions are really helpful. I would also like to remove theme, textScaleFactor, hideCurrentSnackBar , scaffoldBackgroundColor, mediaQueryViewPadding, primaryColor but they are used in code at the moment.

Okay, fine with me, if you think it's better this way.

Jonas-Sander (approved):

PR #439, opened 2023-03-08
Add push notifications support for Android >= 13 and macOS (limited)

### LINKED ISSUE #262: Request push notification permission for Android 13
Android 13 requires apps to request permission to send push notifications, similar as iOS does it. Therefore, we need to adjust the `requestIOSPermissions` method and add the ask for push notifications permissions screen into the onboarding for Android 13 devices.

https://github.com/SharezoneApp/sharezone-app/blob/769a8bdbe31178e646b4ad9dc4fc38db59c00884/app/lib/notifications/firebase_messaging_callback_configurator.dart#L128

nilsreichardt:
## Description

* Moving the logic for when we should request the push notifications to one place
* Add support for Android
* Add (limited) support for macOS
  * The problem here is that the `firebase_messaging` package is not official support for macOS. It just uses the Firebase iOS SDK. Therefore, the `firebase_messaging` is unable to ask for push notification permission. Users need to grant the push notification for themselves in the system settings. In the future, we can add introductions for our users, add a warning when the permission, is not granted, etc. However, I wouldn't do this now because at the moment we have more important tickets.
  * The application doesn't need to be open to receive push notifications 👍
* The PR also adds the required configurations for web (using VAPID key). However, because of https://github.com/firebase/flutterfire/issues/10870 we do not enable push notifications on the web yet. 

### Android

telegram-cloud-photo-size-2-5359645785978947373-y

### macOS

https://user-images.githubusercontent.com/24459435/235187914-e5a20e52-d888-4af9-953b-73cc9751a19d.mov

## Related Tickets

Closes #262


nilsreichardt:
@Jonas-Sander Is the structure okay?

`NotificationsPermissionBloc` is not really bloc. Should we still keep the suffix of "Bloc"?

nilsreichardt:
Status: Unit & manual tests are missing

nilsreichardt:
Oh, I have already the unit tests, so manual testing is last thing that is missing

PR #441, opened 2023-03-08
Fix trim "What's new" string for publishing the alpha version to TestFlight v2

nilsreichardt:
._.


nilsreichardt:
Since we reverted the trimming the "What's new" string (#442), we can close this PR.

PR #445, opened 2023-03-09
Migrate `authentication_base` package to null safety

### LINKED ISSUE #207: Migrate the `authentification_base` package to null safety
Blocked by:
* #210 
* #206 
* #187 
* #186 
* #182

nilsreichardt:
Closes #207


lib/authentification/authentification_base/lib/src/models/demo_user.dart
```diff
@@ -6,9 +6,19 @@
...
+
 import 'package:authentification_base/authentification.dart';
+import 'package:firebase_auth/firebase_auth.dart';
+import 'package:firebase_auth_platform_interface/src/auth_provider.dart';
 
 class DemoUser extends AuthUser {
+  DemoUser()
+      : super(
+          uid: 'testUser',
+          firebaseUser: DemoFirebaseUser(),
```
nilsreichardt, line 19: Not sure how we should handle this the `DemoUser`. Because we are forced to provider a `firebaseUser` or restructure the user (like making the `firebaseUser` null). The `User` of the Firebase Auth package has no constructor so we need to create our own `DemoFirebaseUser` which brings the problem of implementing methods like `delete()`

lib/authentification/authentification_base/lib/src/apple/apple_sign_in_logic.dart
```diff
@@ -13,27 +13,20 @@ import 'package:sharezone_utils/platform.dart';
 import 'package:sign_in_with_apple/sign_in_with_apple.dart';
 
 class AppleSignInLogic {
-  Future<User> signIn() async {
+  Future<User?> signIn() async {
```
Jonas-Sander, line 16: Shouldn't this throw an `Exception` or return a user?

Jonas-Sander (approved):

PR #448, opened 2023-03-09
Fix "sz pub get" not running for all packages.

Jonas-Sander:


tools/sz_repo_cli/lib/src/commands/src/pub_get_command.dart
```diff
@@ -55,9 +55,7 @@ class PubGetCommand extends Command {
 
     final res = taskRunner
         .runTaskForPackages(
-          packageStream: repo
-              .streamPackages()
-              .where((package) => package.hasTestDirectory),
```
nilsreichardt, line 60: lol

nilsreichardt (approved):
LGTM

tools/sz_repo_cli/lib/src/commands/src/pub_get_command.dart
Jonas-Sander: I copied it from the test_command code I think and forgot to remove this 🤡  
I mean should've been catched by the reviewer of the PR where I added this obviously ;P

nilsreichardt (approved):
LGTM

(maybe it's smarter for the next time to split this PR into two so that the fix for CI doesn't need to run through the app CI)

Jonas-Sander:
> LGTM
> 
> (maybe it's smarter for the next time to split this PR into two so that the fix for CI doesn't need to run through the app CI)

Yeah hadn't thought about that 😅

PR #449, opened 2023-03-09
Delete old unused code in `pub_get_command.dart`.

Jonas-Sander:
This code was replaced by the more generalized code in `concurrent_package_task_runner.dart`.


nilsreichardt (approved):
LGTM

PR #453, opened 2023-03-09
Skip `web-preview` workflow for merge queues

nilsreichardt:
Running the `web-preview` workflow for non `pull_request` types doesn't make sense because this action needs to comment to a pull request. With this change, we avoid running the workflow in merge queues.


.github/workflows/app_ci.yml
```diff
@@ -305,6 +305,9 @@ jobs:
     permissions:
       pull-requests: write # for FirebaseExtended/action-hosting-deploy to comment on PRs
       checks: write # for FirebaseExtended/action-hosting-deploy to comment on PRs (without write permissions for checks the action doesn't post a comment to the PR, we don't know why)
+    # We only want to build the web app for PRs because other types doesn't
+    # make sense.
```
Jonas-Sander, line 309:
```suggestion
    # We only want to build the web app only for PRs.
    # Otherwise this will be triggered inside a merge-queue.
```

PR #469, opened 2023-03-10
Migrate the `user` package to null safety

### LINKED ISSUE #182: Migrate the `user` package to null safety
Blocked by:
* #184 
* #187 
* #201

nilsreichardt:
Migrated to null safety. Removed `quiver` because it would require that the `sharezone` package would upgrade to Dart 2.12 (the package was only used for a the `hashCode` in `Period`). 

Tested it manually.

Closes #182


lib/user/lib/src/models/timetable/period.dart
```diff
@@ -82,54 +82,52 @@ class Period {
...
   }) {
     return Period(
       number: number,
       startTime: startTime ?? this.startTime,
       endTime: endTime ?? this.endTime,
     );
   }
+
+  @override
+  int get hashCode => number.hashCode ^ startTime.hashCode ^ endTime.hashCode;
```
Jonas-Sander, line 122:
```suggestion
  int get hashCode => Object.hash(number, startTime, endTime);
```

btw, since some time you can write this instead

lib/user/lib/src/models/user.dart
```diff
@@ -22,41 +21,41 @@ class AppUser {
...
+    required this.userTipData,
+    required this.createdOn,
     this.features,
   });
 
-  factory AppUser.create({String id}) {
+  factory AppUser.create({String? id}) {
     return AppUser._(
-      id: id,
+      id: id ?? "Id",
```
Jonas-Sander, line 58: I don't think `id` should be nullable. I see that `AppUser.create` was used in several places as for a placeholder value (and thus didn't pass `id`) but I think the right thing to do is to make `id` required and non-nullable and just pass a placeholder id in the aforementioned places.

Jonas-Sander (approved):

PR #478, opened 2023-03-10
Use fix versions for tools and Actions in our CI

nilsreichardt:
To follow [security best practices](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) and to have a deterministic setup, we fix our version for tools and Actions that we are using in our CI. 

This PR updates also some Actions because some Actions haven't published an update for months but committed new changes.

### DIFF
--- a/.github/workflows/alpha.yml
+++ b/.github/workflows/alpha.yml
@@ -89,10 +89,10 @@ jobs:
     # So we can just use "sz COMMAND" instead of "dart ../path/to/script.dart ..."
     - run: echo $(realpath ./bin) >> $GITHUB_PATH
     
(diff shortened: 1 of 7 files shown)


Jonas-Sander (approved):
Very nice.

Jonas-Sander:
There is one problem though. When calling `sz analyze` this will run the following code which always activates the latest `tuneup` version used for analysis:
```
    await runProcessSucessfullyOrThrow(
        'fvm', ['dart', 'pub', 'global', 'activate', 'tuneup'],
        workingDirectory: repo.sharezoneFlutterApp.location.path);
```

PR #495, opened 2023-03-12
Migrate `group_domain_models` package null safety

### LINKED ISSUE #192: Migrate the `group_domain_models` package to null safety

nilsreichardt:
Closes #192


lib/group_domain_models/lib/src/models/course.dart
```diff
@@ -7,39 +7,39 @@
...
 import 'group_info.dart';
 import 'group_type.dart';
-import 'group_key.dart';
+import 'member_role.dart';
+import 'write_permissions.dart';
 
 class Course {
-  final String id, name, subject, abbreviation, sharecode, joinLink, meetingID;
-  final MemberRole myRole;
+  final String? id, name, subject, abbreviation, sharecode, joinLink, meetingID;
```
Jonas-Sander, line 20: Can't you make `id` non-nullable?

lib/group_domain_models/lib/src/models/course.dart
```diff
@@ -166,32 +166,28 @@ class Course {
...
-  GroupKey getGroupKey() {
-    return GroupKey(id: id, groupType: GroupType.course);
-  }
 }
 
 class CourseData {
-  final String id, name, subject, description, abbreviation;
-  final String sharecode, joinLink, referenceSchoolID, meetingID;
-  final List<String> referenceSchoolClassIDs;
+  final String? id, name, subject, description, abbreviation;
```
Jonas-Sander, line 172: Can't you make `id` non-nullable?

lib/group_domain_models/lib/src/models/course.dart
```diff
@@ -307,12 +305,12 @@ class CourseData {
 class CourseSettings {
   final bool isPublic;
   final bool isMeetingEnabled;
-  final WritePermission writePermission;
+  final WritePermission? writePermission;
```
Jonas-Sander, line 308: Couldn't this be non-nullable and be the most restrictive option if none are given? Or is this on purpose because e.g. the backend assigns it and we don't want to set it?  

I'm just thinking that if some logic needs this value but its `null` then the only sensible option would be to interpret this as the most restrictive WritePermissions anyways, so we could make it non-nullable and implement it here.  
Idk if this makes sense, I'm not too familiar with this code.

lib/group_domain_models/lib/src/models/school.dart
```diff
@@ -6,30 +6,31 @@
...
 // SPDX-License-Identifier: EUPL-1.2
 
-import 'package:meta/meta.dart';
 import 'course.dart';
 import 'member_role.dart';
 
 class School {
-  final String id, name, publicKey, joinLink;
-  final MemberRole myRole;
+  final String? id, name, publicKey, joinLink;
```
Jonas-Sander, line 13: Can't you make `id` non-nullable? (Are we even using `School`? I don't think so, right?)

lib/group_domain_models/lib/src/models/school.dart
```diff
@@ -38,19 +39,19 @@ class School {
 }
 
 class SchoolData {
-  final String id, name, title, description, abbreviation;
-  final String publicKey, joinLink;
+  final String? id, name, title, description, abbreviation;
```
Jonas-Sander, line 42: Can't you make id non-nullable? (Are we even using SchoolData? I don't think so, right?)

lib/group_domain_models/lib/src/models/school_class.dart
```diff
@@ -8,39 +8,40 @@
...
-import 'group_key.dart';
+import 'group_info.dart';
+import 'group_type.dart';
 import 'member_role.dart';
 
 class SchoolClass {
-  final String id, name, sharecode, joinLink, meetingID;
-  GroupId get groupId => GroupId(id);
-  final MemberRole myRole;
+  final String? id, name, sharecode, joinLink, meetingID;
```
Jonas-Sander, line 19: Can't you make id non-nullable?

lib/group_domain_models/lib/src/models/school_class.dart
```diff
@@ -104,28 +105,24 @@ class SchoolClass {
...
-
-  GroupKey getGroupKey() {
-    return GroupKey(id: id, groupType: GroupType.schoolclass);
-  }
 }
 
 class SchoolClassData {
-  final String id, name, description, abbreviation, meetingID;
-  final String sharecode, joinLink, referenceSchoolID;
+  final String? id, name, description, abbreviation, meetingID;
```
Jonas-Sander, line 111: Can't you make id non-nullable?

lib/group_domain_models/lib/src/models/school_class.dart
```diff
@@ -142,8 +139,10 @@ class SchoolClassData {
     );
   }
 
-  factory SchoolClassData.fromData(
-      {@required String id, @required Map<String, dynamic> data}) {
+  factory SchoolClassData.fromData({
+    required String id,
+    required Map<String, dynamic> data,
```
Jonas-Sander, line 144: Nit (maybe for some other PR): For some classes `data` is positional and `id` is named and for example here both is named. We should decide on one of both for all cases.

lib/group_domain_models/lib/src/models/write_permissions.dart
```diff
@@ -6,18 +6,15 @@
...
-    case WritePermission.everyone: return 'Alle';
-    case WritePermission.onlyAdmins: return 'Nur Admins';
-    default: return 'Einstellung konnte nicht gefunden werden';
+    case WritePermission.everyone:
+      return 'Alle';
+    case WritePermission.onlyAdmins:
+      return 'Nur Admins';
+    default:
+      return 'Einstellung konnte nicht gefunden werden';
   }
```
Jonas-Sander, line 19: Couldn't we only use both cases in the switch and put the fallback as a separate return after it so that if a new `WritePermission` case is introduced we get a warning here?

Jonas-Sander (changes requested):

lib/group_domain_models/lib/src/models/course.dart
Jonas-Sander:
```dart
  factory Course.create() {
    return Course._(
      id: null,
      name: "",
      subject: "",
```

Since e.g. in here you could just replace `null` with a placeholder value if you need to.

Jonas-Sander (approved):
@nilsreichardt Do you want to look at it again or should I go ahead and merge?

nilsreichardt (commented):
LGTM

PR #504, opened 2023-03-13
Create and use `ConcurrentCommand` for sz cli.

Jonas-Sander:
By using the `ConcurrentCommand` class, commands can be created using less code to run a task for several packages concurrently.


tools/sz_repo_cli/lib/src/common/src/concurrent_command.dart
```diff
@@ -0,0 +1,102 @@
...
+  ///
+  /// This can be overriden by the user via command line argument.
+  Duration get defaultPackageTimeout => Duration(minutes: 10);
+
+  /// Number of packages that are going to be processed concurrently.
+  ///
+  /// Setting this to <= 0 means that there is no concurrency limit, i.e.
+  /// everything will be processed at once.
+  ///
+  /// This can be overriden by the user via command line argument.
```
nilsreichardt, line 40:
```suggestion
  /// This can be overridden by the user via command line argument.
```

tools/sz_repo_cli/lib/src/common/src/concurrent_command.dart
```diff
@@ -0,0 +1,102 @@
...
+
+  @override
+  Future<void> run() async {
+    isVerbose = argResults['verbose'] ?? false;
+    await runSetup();
+
+    final _max = argResults[maxConcurrentPackagesOptionName];
+    final maxNumberOfPackagesBeingProcessedConcurrently = _max != null
+        ? int.tryParse(argResults[maxConcurrentPackagesOptionName])
+        // null as interpreted as "no conucrrency limit" (everything at once).
```
nilsreichardt, line 71:
```suggestion
        // null as interpreted as "no concurrency limit" (everything at once).
```

nilsreichardt:
@Jonas-Sander Had the problem that the `sharezone` package timed out also a few times. Mabe it's something similar like in #503 where it happens in 5% of the cases.

.github/workflows/app_ci.yml
```diff
@@ -145,7 +145,7 @@ jobs:
       - run: echo $(pwd)/bin >> $GITHUB_PATH
       
       - name: Run tests via "sz test"
-        run: sz test -c 4
+        run: sz test -c 4 --package-timeout-minutes 15
```
nilsreichardt, line 148: You changed the app CI but the CLI CI failed 😅

PR #505, opened 2023-03-13
Increase test timeout to 15 minutes.

Jonas-Sander:
In #504 running `sz test` timed out for the `sharezone` package. From what I can see it doesn't have something to do with the PR in #504 itself, so I increase the timeout from 10 to 15 minutes.

### DIFF
--- a/.github/workflows/app_ci.yml
+++ b/.github/workflows/app_ci.yml
@@ -145,8 +145,7 @@ jobs:
       - run: echo $(pwd)/bin >> $GITHUB_PATH
       
       - name: Run tests via "sz test"
-        run: sz test -c 4
-      
+        run: sz test -c 4 --package-timeout-minutes 15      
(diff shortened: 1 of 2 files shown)


nilsreichardt:
@Jonas-Sander I would recommend to first merge #504 and then this PR because otherwise you need to wait ~45 minutes until you have #504 in the `main` branch because this PR also executes the integration tests

Jonas-Sander:
> @Jonas-Sander I would recommend to first merge #504 and then this PR because otherwise you need to wait ~45 minutes until you have #504 in the `main` branch because this PR also executes the integration tests

So should I force-merge the other PR even if the `test` job fails again?

nilsreichardt:
> So should I force-merge the other PR even if the test job fails again?

No, but the `test` job will finish in the next minutes because the timeout is flaky. Re-trigger the CI is enough

PR #533, opened 2023-03-28
Use `sharezone_lints` in `notifications`.

Jonas-Sander:


nilsreichardt (commented):
Where is the `sharezone_lints` package?

PR #534, opened 2023-03-28
Use `sharezone_lints` in `sharezone_widgets`.

Jonas-Sander:


nilsreichardt (commented):
Not sure if here is everything right because this PR has also some changes in `app` 🤔

lib/sharezone_widgets/lib/src/widgets.dart
```diff
@@ -171,8 +173,8 @@ class DeleteButtonWithPoppingTrue extends StatelessWidget {
 
 enum LogoColor {
   white,
-  blue_long,
-  blue_short,
+  blueLong,
+  blueShort,
 }
```
Jonas-Sander, line 179: @nilsreichardt Yes since I had to change this enum because of a lint, I needed to update its usages inside `app` as well.

nilsreichardt: makes sense haha

nilsreichardt (approved):
LGTM

PR #536, opened 2023-03-28
Fix lints in `sharezone-app`.

Jonas-Sander:
Fix lints that appeared when migrating several packages to `sharezone_lints`. The curious thing is that this PR doesn't change `app` to use `sharezone_lints`. So my guess is that the Dart analyzer had some hickups. As this is stuff we have to change anyways in the future I would just go forward with these changes.


nilsreichardt (approved):
LGTM

PR #537, opened 2023-03-30
Fix typo in authentification label in `labeler.yml`

nilsreichardt:


.github/labeler.yml
```diff
@@ -140,7 +140,7 @@
   - "**/attachments/**"
   - "**/**attachment**"
 
-"feature: authentication":
+"feature: authentification":
```
Jonas-Sander, line 143: `authentication` seems to be the preferable English word: https://english.stackexchange.com/questions/7844/is-authentification-a-real-word

nilsreichardt: yeah but currently we have this everywhere wrong (including the label) 😅 should we rename the label?

Jonas-Sander: oh... yeah i guess 😅

but then lets merge this first so that for now there is consistency.

PR #540, opened 2023-03-31
Remove donation code.

Jonas-Sander:


nilsreichardt (approved):
LGTM

(I think we need to add RevenueCat again on Monday when we start to work on the subscriptions but that's okay)

Issue #480, opened 2023-03-10
Replace `package:tuneup` used in the Sharezone CLI.


Jonas-Sander:
Since `package:tuneup` [has been discontinued](https://github.com/google/tuneup.dart/issues/96) we should replace it.  
We can use `flutter analyze` and `dart analyze` instead, but neither have `--fail-on-todos` flag which we use.  
We might have to implement a seperate "fail on TODOs" step ourselves.

See also [this comment](https://github.com/SharezoneApp/sharezone-app/pull/478#issuecomment-1464468646) on #478

nilsreichardt:
What if we make `TODO`s to warnings in our `analysis_options.yaml`?:

```yaml
analyzer:
  errors:
    todo: warning
```

Having TODOs and running `dart analyze` returns exit code 2.

Jonas-Sander:
Daaammmnnn you smaart

PR #541, opened 2023-03-31
Treat TODOs as warnings.

Jonas-Sander:
By treating TODOs as warnings we can use the normal Dart analyzer instead of using the deprecated `package:tuneup` to run `sz analyze` and fail if TODOs are found (see https://github.com/SharezoneApp/sharezone-app/issues/480).


nilsreichardt (approved):
LGTM

Issue #545, opened 2023-03-31
Workflow "macos-build-test" is not working when a PR is opened from a fork 


Jonas-Sander:

Jonas-Sander:
Related to #348

Issue #296, opened 2022-09-27
Workflow "ios-integration-test" is not working when a PR is opened from a fork


nilsreichardt:

nilsreichardt:
Related to #348

Issue #295, opened 2022-09-27
Workflow "web-preview" is not working when a PR is opened from a fork


nilsreichardt:

nilsreichardt:
Related to #348

PR #559, opened 2023-04-04
Require fork PRs to have a safe to test label before running the tests.

### LINKED ISSUE #348: Some GitHub Actions will always fail for external PRs.
When a PR is opened by an external contributor the actions `web-preview` and `ios-integration-test` will always fail since they can't access secrets.   
Example PR: https://github.com/SharezoneApp/sharezone-app/pull/347

As described in [this article](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) Actions that have the trigger `pull_request` can't access secrets, if they are triggered on a PR from an external contributor (no write permissions for the repo), so that these secrets can't be easily stolen.  

Because we use these secrets for the following tasks, these tasks will fail if the secrets are not available
1. Deploy the app (`web-preview` action)
  Secret: `FIREBASE_HOSTING_PROD_KEY`
2. Run integration tests with credentials for accounts (`ios-integration-test`, `android-integration-test`)
  Secrets: `INTEGRATION_TEST_USER_1_EMAIL`, `INTEGRATION_TEST_USER_1_PASSWORD`
3. Build MacOS app (`macos-build-test`)
https://github.com/SharezoneApp/sharezone-app/actions/runs/4583590526/jobs/8094642454

### `web-preview` action

Split the workflow into building the code for web via a `pull_request` trigger, upload the artifact and run a second action on success via the `workflow_run` trigger, which will upload the builded code to Firebase Hosting. This would seperate building the untrusted code (dangerous) and uploading the code that was just build (not dangerous - hopefully 😉).  
The second Action will have access to the secrets, since it is triggered on `workflow_run`, not `pull_request`.   
This approach is described in more details with examples [here](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).

### `ios-integration-test`, `android-integration-test` action

Since we can't split the workflow into several parts like above we need another approach.  

Instead of manually creating account credentials for the tests to use, we could automate creating new Sharezone accounts via the test code. This would not only include creating a Firebase Auth account, but also adding the necessary setup data into the user document (e.g. type of user - student, teacher, parent).

nilsreichardt:
## Description

_Warning:_ I'm uncertain if everything works because we are now using a `pull_request_target` trigger which uses the code from our `main` branch. So I'm only able to fully test the pull request when we merged this PR. I'm not sure if is going to break in our merge queue.

This is the flow for being able to run securely our CI checks that also need to access our secrets:

```mermaid
graph TD
A{Is PR a fork?}
A -- Yes --> B[Remove the 'safe to build' label]
A -- No  --> C[Execute the 'changes' workflow]
B --> C
C --> D{Is PR a fork?}
D -- Yes --> E{Has 'safe to build' label?}
D -- No  --> H[Execute the workflows]
E -- Yes --> H
E -- No  --> F[Cancel]
```

## Related Tickets

Closes #348
Closes https://github.com/SharezoneApp/sharezone-app/issues/295
Closes https://github.com/SharezoneApp/sharezone-app/issues/296
Closes https://github.com/SharezoneApp/sharezone-app/issues/544
Closes https://github.com/SharezoneApp/sharezone-app/issues/545


Jonas-Sander (changes requested):
Edit: Sorry got the code wrong. You already implemented it, right?

I think the label needs to be automatically removed after a new commit was made. Otherwise an attacker can just make a "good" PR, wait for label to be assigned and push malicious code afterwards.

nilsreichardt:
> You already implemented it, right?

I'm testing it now if it's working for non fork PRs. For Fork PRs we can only test it really when PRs has been merged. However, I tested this part already in my GitHub Actions ([remove-safe-to-test-label](https://github.com/nilsreichardt/remove-safe-to-test-label) and [verfiy-safe-to-test-abel](https://github.com/nilsreichardt/verfiy-safe-to-test-label)).

>  think the label needs to be automatically removed after a new commit was made. Otherwise an attacker can just make a "good" PR, wait for label to be assigned and push malicious code afterwards.

This is the case because the "remove-safe-to-build-label" job is executed as first job. So this jobs removes the "safe-to-test" label. The jobs that are coming later, would fail because the "safe to test" label is missing.

Jonas-Sander (changes requested):
* uses: nilsreichardt/remove-safe-to-test-label move these actions to SharezoneApp org+
* Split files into actions with `pull_request_target` trigger and `pull_request` trigger ("unsafe" and "safe" jobs)

Jonas-Sander (approved):
LGTM

nilsreichardt:
I need to force merge this because this PR uses the `pull_request_target` trigger and the moment is not `pull_request_target` for this file on the `main` branch. Therefore, the checks are never executed.

PR #571, opened 2023-04-05
Add option to clear app cache

### LINKED ISSUE #570: Add a hidden feature to clear the app cache
For iOS users, it's not easy to delete the app cache. To debug bugs like [this one](https://discord.com/channels/545668860859252746/616181222577668106/1093247845810847847) it would great if users could delete the app cache.

nilsreichardt:
## Description

Users can clear the app cache with the "ClearCache" activation code.

https://user-images.githubusercontent.com/24459435/230214220-9c523a51-3693-4f4f-91c8-28fb5b4d372c.mov

## Related Tickets

Closes #570

### DIFF
--- a/app/lib/activation_code/src/bloc/enter_activation_code_bloc.dart
+++ b/app/lib/activation_code/src/bloc/enter_activation_code_bloc.dart
@@ -10,8 +10,11 @@ import 'package:analytics/analytics.dart';
 import 'package:app_functions/sharezone_app_functions.dart';
 import 'package:bloc_base/bloc_base.dart';
 import 'package:crash_analytics/crash_analytics.dart';
+import 'package:feature_discovery/feature_discovery.dart';
+import 'package:flutter/material.dart';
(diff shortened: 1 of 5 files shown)


Jonas-Sander (approved):
LGTM

Does this also clear all preferences from `StreamingKeyValueStore`?  
Also in the dev option we had this code as well:
```dart
FeatureDiscovery.clearPreferences(context, [
  blackboardItemReadByUsersListFeatureDiscoveryStepId,
]);
```
don't know if we would want to include here as well, but just as an info.

nilsreichardt:
> Does this also clear all preferences from StreamingKeyValueStore?

Yes, `streaming_shared_preferenes` uses `shared_preferences` under the hood 👍 

> FeatureDiscovery.clearPreferences(context, [

Would be good, but this requires the `BuildContext` in the bloc. How should we solve this?

> Ah btw what do you think of calling .toLowerCase() on the input as well? might make it easier to insert the codes

Yes, we can do this 👍

PR #610, opened 2023-04-06
Add Sharezone Plus feature guard and guard some features.

Jonas-Sander:
Added `SharezonePlusFeatureGuard` widget and guarded:
* viewing submissions list
* viewing homework done by list
* information sheet read by list

The guard should only be considered a basic implementation as we don't know yet how accessing Sharezone Plus features without a subscription should look like in the final version (it might very well differ from feature to feature).  
My guess is that we might add code in the future that navigates away or shows a dialog (or does something custom for different features) even before we reach the guard (i.e. navigate to the page with the plus feature). 
This guard might still be in place then anyways as users might try to reach these pages by changing URLs or in some other way.

https://user-images.githubusercontent.com/29028262/230422622-e3044343-a788-4ec9-9c30-386e07682630.mp4


nilsreichardt (approved):
LGTM

PR #612, opened 2023-04-06
Add `sz format` command.

### LINKED ISSUE #589: Add command to format code, like `sz format`
`sz format` formats our code with our formatting standards. At the moment, it's just `dart format .`

Jonas-Sander:
Closes #589

### DIFF
--- a/tools/sz_repo_cli/lib/src/commands/src/format_command.dart
+++ b/tools/sz_repo_cli/lib/src/commands/src/format_command.dart
@@ -1,0 +1,30 @@
+// Copyright (c) 2022 Sharezone UG (haftungsbeschränkt)
+// Licensed under the EUPL-1.2-or-later.
+//
+// You may obtain a copy of the Licence at:
+// https://joinup.ec.europa.eu/software/page/eupl
+//
+// SPDX-License-Identifier: EUPL-1.2
+
+import 'dart:async';
+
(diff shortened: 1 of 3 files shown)


nilsreichardt (approved):
LGTM

(can we use `sz format` inside `sz analyze`?)

Jonas-Sander:
> 

What do you mean exactly?

nilsreichardt:
> What do you mean exactly?

My thought was that have now two places where we define how to format code, in `sz format` and `sz analyze`. When we change how we format code, we would need to update it at two places. But maybe this too overkill for now because I don't we going to update that often.

But maybe a comment above `return runProcessSucessfullyOrThrow('fvm', ['dart', 'format', '.'],` that if change this command, we also need to change it in `sz analyze`

Jonas-Sander:
> > What do you mean exactly?
> 
> My thought was that have now two places where we define how to format code, in `sz format` and `sz analyze`. When we change how we format code, we would need to update it at two places. But maybe this too overkill for now because I don't we going to update that often.
> 
> But maybe a comment above `return runProcessSucessfullyOrThrow('fvm', ['dart', 'format', '.'],` that if change this command, we also need to change it in `sz analyze`, would be good

I'm just gonna let both use one format method with an argument, after both PRs have been merged 👍

PR #614, opened 2023-04-06
Add `--exclude-goldens` and `--only-goldens` argument to `sz test`

nilsreichardt:
Adds a `--exclude-goldens` and `--only-goldens` argument to the `sz test` command. By default, you run normal test and golden tests. With the arguments you can decide what do you want to run.

As of now we need to put our golden tests into the `test_goldens` folder (instead of the `test` folder).


Jonas-Sander (approved):
LGTM

PR #568, opened 2023-04-05
Fix picking the homework reminder minutes.

Jonas-Sander:
On Android picking only reminder times in an half-hour interval was broken. Instead a user could pick any time. By using `package:interval_time_picker` this issue is now fixed an users can again only pick reminder times in an half-hour interval.

grafik

Fixes #565

PR #616, opened 2023-04-06
Fix setting minutes in time pickers

### LINKED ISSUE #584: It's only possible to set the minute of an event to 0 or 30
#568 introduced a new bug: When adding a new event (e.g. an exam) it's only allowed to enter 0 or 30 as value for the minute at the time picker.

https://user-images.githubusercontent.com/24459435/230375636-09036a82-455a-4acb-8831-fc3da3c5f5dd.mov

nilsreichardt:
## Description

#568 introduced a new bug: When adding a new event (e.g. an exam) it's only allowed to enter 0 or 30 as value for the minute at the time picker.

https://user-images.githubusercontent.com/24459435/230375636-09036a82-455a-4acb-8831-fc3da3c5f5dd.mov

The problem was the minute interval for the Material time picker was hard-coded.

## Solution demo

https://user-images.githubusercontent.com/24459435/230507536-e35099d9-c0d2-4842-9bec-95f99d5ceb13.mov

## Related Tickets

Closes #584


app/lib/timetable/src/edit_time.dart
```diff
@@ -114,8 +115,9 @@ Future<Time> selectTime(BuildContext context,
   return showIntervalTimePicker(
     context: context,
     initialTime: initialTime?.toTimeOfDay() ?? TimeOfDay(hour: 9, minute: 10),
-    interval: 30,
-    visibleStep: VisibleStep.thirtieths,
+    interval: minutesInterval,
+    visibleStep:
+        minutesInterval == 30 ? VisibleStep.thirtieths : VisibleStep.fifths,
```
Jonas-Sander, line 120: If we're basically hard coding the logic here that it's either 30 or 5 minute intervals then shouldn't we also also add an assert that `minutesInteval` is one or the other?

Jonas-Sander (approved):
LGTM, see comment

app/lib/timetable/src/edit_time.dart
nilsreichardt: I'm using an extension now on `int` to get the `VisibleStep` and throw an exception if it's an unknown minute value:

```dart
extension on int {
  VisibleStep toVisibleStep() {
    switch (this) {
      case 1:
      case 5:
        return VisibleStep.fifths;
      case 6:
        return VisibleStep.sixths;
      case 10:
        return VisibleStep.tenths;
      case 20:
        return VisibleStep.twentieths;
      case 30:
        return VisibleStep.thirtieths;
      case 60:
        return VisibleStep.sixtieth;
      default:
        // At the moment, only these intervals are supported. If you need
        // another one, please add it and handle it in the switch statement.
        throw Exception(
          "Unsupported minutes interval: $this. Please handle the other cases yourself.",
        );
    }
  }
}
```

PR #618, opened 2023-04-07
`sz` CLI: Move logic out of `package.dart`.

Jonas-Sander:
Instead of letting `package.dart` grow and grow, I moved all of the logic out to the `xxx_command.dart` classes.   
Additionally I removed the `DartPackage` and `FlutterPackage` subclasses as they are unnecessary with the logic moved elsewhere.


nilsreichardt (approved):
LGTM

PR #623, opened 2023-04-16
Fix test commands in CI

nilsreichardt:


Jonas-Sander (approved):
Ah man what a nasty bug :D

PR #79, opened 2021-12-26
Replace privacy policy page.

Jonas-Sander:
Replace old privacy policy pages with a new markdown-based, multi-layout privacy policy page.

This PR will **not** update the privacy policy itself, only the presentation.   

Note: This PR originally had code for loading the privacy policy asynchronously (it wasn't loaded from the internet, but all the code for loading indicators, errors, etc. was 90% there). I removed it for now, to have less complex code as we don't need it yet. The deleted code: [7eb67fb3fd4694e1cf4505cc12df821198ea7f22](https://github.com/SharezoneApp/sharezone-app/pull/79/commits/a43badb0171c544d4fc9944bb1000ae95685240e)

**Desktop / Narrow layout**
(Updated: "Weitere Optionen" in the lower-left is now centered)

https://user-images.githubusercontent.com/29028262/231279789-27777b7e-1ac2-4bd1-9fba-f5c57a19ba48.mp4

**Mobile layout**

| Onboarding | Default |
|-|-|
| grafik |grafik |

| Dark mode | Light mode |
|-|-|
| grafik |grafik |
| grafik |grafik|
| grafik |grafik|

**Privacy policy with Subchapters** (just for demonstration purposes)

https://user-images.githubusercontent.com/29028262/236900647-93ae287c-eb70-42bc-8a53-12cb2d267db0.mp4

Tidbits to remeber:
* Expansion behavior is different on layouts with bottom sheets in contrast to the desktop layout with the fixed table of contents on the side. On the latter manually expanded sections will stay expanded until they are manually collapsed again, on the former they will always automatically collapse again as soon as they are not read anymore.  

TODOs:
- [x] A section should already be marked as "currently reading" when the heading is at the top of the page not only after we scrolled past it (because if we use the TOC to navigate to a section it will position the heading at the top of the page - also its just nicer in general).
- [x] Fix Anchor for "11. Verarbeitung des gewählten Account-Typs und des Bundeslandes" not working (maybe more). The generated ID by the Flutter `markdown` plugin is different from the Id we use in the text. Seems to happen for Anchors with Sonderzeichen/Umlaute?
- [x] `flutter: 'package:sharezone/onboarding/sign_up/pages/privacy_policy/src/table_of_contents/currently_reading.dart': Failed assertion: line 104 pos 14: 'oldVisibleHeadings.length == 1': is not true.`
[...]

PR #624, opened 2023-04-19
Add `VisualDensitySetting` class.

Jonas-Sander:
This class is a prerequisite for #79.

With this class we can track if the user has manually chosen a specific `VisualDensity` or if they use the default `VisualDensity.adaptivePlatformDensity` (one can't tell by looking just at the `VisualDensity` class).

### DIFF
--- a/app/lib/account/theme/theme_settings.dart
+++ b/app/lib/account/theme/theme_settings.dart
@@ -86,7 +86,7 @@ class ThemeSettings extends ChangeNotifier {
     required double defaultTextScalingFactor,
 
     /// The value assigned to [visualDensity] if no other value is cached.
-    required VisualDensity defaultVisualDensity,
+    required VisualDensitySetting defaultVisualDensity,
 
     /// The value assigned to [themeBrightness] if no other value is cached.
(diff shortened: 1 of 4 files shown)


app/lib/account/theme/theme_settings.dart
```diff
@@ -121,20 +121,20 @@ class ThemeSettings extends ChangeNotifier {
...
       name: 'ui_visual_density_changed',
       data: {
-        'visual_density': {
-          'horizontal': value.horizontal,
-          'vertical': value.vertical
-        }
+        // bool is not allowed in firebase_analytics so we use `toString`.
+        'isAdaptivePlatformDensity': value.isAdaptivePlatformDensity.toString(),
+        'horizontal': value.visualDensity.horizontal,
+        'vertical': value.visualDensity.vertical
```
Jonas-Sander, line 137: The old code would have always thrown an error here since the map inside map (`'data:': { 'visual_density': {...}}`) is not allowed by `firebase_analytics`.

nilsreichardt (approved):
LGTM

PR #653, opened 2023-05-08
Change "Akzeptieren" to "Weiter" in onboarding privacy policy step.

Jonas-Sander:
As "Akzeptieren" is wrong for a privacy policy and might get us into legal trouble, we replace it with "Weiter".

https://easyrechtssicher.de/ich-habe-die-datenschutzerklaerung-gelesen-und-akzeptiere-diese/


nilsreichardt (approved):
LGTM

PR #655, opened 2023-05-09
Fix loading homeworks for teachers

### LINKED ISSUE #497: App freezes when signing out
## Describe the bug

When I'm signing out, freezes the app.

## Steps to reproduce the bug

1. Sign in with your Sharezone account
2. Sign out

## Current _broken_ behavior

App freezes. It's impossible to press something.

## Expected behavior

I should be able to log out.

## Screenshots, videos or logs

https://user-images.githubusercontent.com/24459435/224519590-49b3b5d2-2f5a-48a9-9a86-1645662c453a.mov

You can see in the video that I signed out. However, after the sign-out I can't press anything anymore. The app is frozen.

## Tested device

Tested with web, macOS and iOS.

- App version: 06d56f79d0303680fc26a514755effc39b4d5ac1

### LINKED ISSUE #608: When signing out, `typeOfUser` is null for a short moment
See https://github.com/SharezoneApp/sharezone-app/pull/573. Maybe this is only on web.

### LINKED ISSUE #654: Homeworks are not loading for teachers
## Steps to reproduce

1. Create a teacher account
2. Add a homework
3. Clear the browser cache
4. Load the web app again

## Demo

image

nilsreichardt:
## Summary

This PR fixes:

* #608 
* #654 
* #497 (real fix)

It's required to this all in one pull request because otherwise it would introduce a new bug.

## What was the issue?

I started to debug on this problem because of #654. I found out it was because `TypeOfUser.student` was used as a backup when it's null (see #573). However, just reverting it to `TypeOfUser.unknown` was not option because this would introduce #497 again.

Therefore, I wondered, why `typeOfUser` can be null. The reason for this was the last time online reporter. Because `.set()` has been used for Firestore, Firestore instantly was setting it null (I think `null` is used for `Firestore.currentTimestamp()` when the client is offline or the server hasn't responded yet). Using `.update()` fixed the issue that the user document only contained `lastOnline`.

But using `update()` in the last online reporter always caused #497. Even using `TypeOfUser.student` or `TypeOfUser.unknown` couldn't be as a workaround. After hours of debugging I found that when signing our the following line was causing the issue:

https://github.com/SharezoneApp/sharezone-app/blob/47e4ccb0993f01b5c9b8c83f31f7ec4c4f833c77/app/lib/util/api/homework_api.dart#L107

The issue is that we listened to `_homeworkSubjectStream` and when an error occurred, we the `.addError` method of `_homeworkSubjectStream ` used. This ended up into an infinity loop which resulted in the freeze (#497),

This was by far one of my hardest bugs... I needed ~5,5h to find the solution and it was extremely hard to debug. 

Fixes #608
Fixes #654
Fixes #497


lib/last_online_reporting/lib/src/implementation.dart
```diff
@@ -24,10 +24,19 @@ class FirestoreLastOnlineReporterBackend {
...
-    );
+    // Do not use .set({merge: true}) here, because the Firestore library sets
+    // 'lastOnline' instantly to null (might be web-only behaviour) which causes
+    // that the user document only contains the 'lastOnline' attribute and
+    // nothing else (missing typeOfUser, etc.).
+    //
+    // Additionally, .update() is here more best practice, set({merge: true})
+    // would be executed even when the user document does not exist (like when
+    // deleting the user of something). This can result into hard bug and side
+    // effects, especially because there is no error. The .update() method would
```
Jonas-Sander, line 35: Hard bug? Do you mean hard to find?

Jonas-Sander (approved):
See comment above

Jonas-Sander:
> This was by far one of my hardest bugs... I needed ~5,5h to find the solution and it was extremely hard to debug.

I think this bug was on me... Sorry!! 🙈😁

Jonas-Sander:
Although in my opinion there should be:
* A lint that warns of piping a stream into itself in this way (at least the error I guess?)
* A debug feature that warns when a stream goes into this kind of infinity loop

If I have time maybe I'm gonna raise an issue on the Dart side, but no promises yet. I guess that there might be reasons why it might be hard to actually implement it as well.

lib/last_online_reporting/lib/src/implementation.dart
nilsreichardt: Yes, hard bugs means bugs that are hard to debug, hard to find, hard to locate, etc

nilsreichardt:
> Although in my opinion there should be:
> 
> * A lint that warns of piping a stream into itself in this way (at least the error I guess?)
> * A debug feature that warns when a stream goes into this kind of infinity loop
> 
> If I have time maybe I'm gonna raise an issue on the Dart side, but no promises yet. I guess that there might be reasons why it might be hard to actually implement it as well.

Definitely, agree. Unfortunately, I couldn't also use performance tools, like DevTools because they're not really working for web and say that I should use Chrome DevTools, however even Chrome DevTools couldn't record it properly.

nilsreichardt:
@Jonas-Sander Any ideas why the android integration test are not passing anymore because of the changes?

Jonas-Sander:
@nilsreichardt Should we release a new version after this is merged?

nilsreichardt:
Yes, makes sense also because of #671

PR #667, opened 2023-05-09
Add docs for Firewall configurations

nilsreichardt:
Some parents ask me via the email support which Firewall configurations are required to get our Sharezone app running. I did some research and used [Charles Proxy](https://www.charlesproxy.com/) to "reverse engineer" our app to find out which domains are used.

### DIFF
--- a/docs/more/firewall.mdx
+++ b/docs/more/firewall.mdx
@@ -1,8 +1,8 @@
 # Firewall-Einstellungen für die Sharezone App
 
-Manche Schule oder Heimnetzwerke schränken den Zugriff auf das Internet ein. Um
-die Sharezone-App korrekt zu nutzen, müssen folgende Domains und Ports in der
-Firewall freigeschaltet werden.
+Manchmal kann es der Fall sein, dass ein Netzwerk den Zugriff auf das Internet
+einschränkt (z.B. Schulen für das Schul-WLAN). Um die Sharezone-App korrekt zu
+nutzen, müssen folgende Domains und Ports in der Firewall freigeschaltet werden.
 
 ## Backend
 
--- a/docs/more/firewall.mdx
+++ b/docs/more/firewall.mdx
@@ -58,4 +58,4 @@ _Ergänzend zu den oben genannten Firewall-Einstellungen._
 
 Falls dennoch Features nicht funktionieren sollten, weil diese durch die
 Firewall geblockt werden, sende uns bitte eine E-Mail mit Details zu dem
-betroffenen Feature an [mailto:support@sharezone.net](support@sharezone.net).
+betroffenen Feature an [support@sharezone.net](mailto:support@sharezone.net).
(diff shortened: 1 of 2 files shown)


docs/more/firewall.mdx
```diff
@@ -0,0 +1,61 @@
...
+- `alt5-mtalk.google.com`
+- `alt6-mtalk.google.com`
+- `alt7-mtalk.google.com`
+- `alt8-mtalk.google.com`
+- `android.apis.google.com`
+- `device-provisioning.googleapis.com` (Sollte bereits mit `*.googleapis.com`
+  abgedeckt sein)
+- `firebaseinstallations.googleapis.com` (Sollte bereits mit `*.googleapis.com` abgedeckt sein)
+
+<!-- Source: https://firebase.google.com/docs/cloud-messaging/concept-options#messaging-ports-and-your-firewall -->
```
nilsreichardt, line 43: Just comment for us where I have these domains from

docs/more/firewall.mdx
```diff
@@ -0,0 +1,61 @@
...
+**Hosts:**
+
+- `*.googleapis.com`
+- `*.sharezone.net`
+- `europe-west1-sharezone-c2bd8.cloudfunctions.net`
+- `*.gstatic.com`
+- `*.app-measurement.com` (Für Google Analytics)
+- `api.revenuecat.com` (Für In-App-Käufe)
+
+**Protokolle:** `HTTP/1.1`
```
Jonas-Sander, line 18: Wouldn't Firestore use WebSockets as well for realtime listeners?

docs/more/firewall.mdx
```diff
@@ -0,0 +1,61 @@
+# Firewall-Einstellungen für Sharezone
+
+Manchmal kann es der Fall sein, dass ein Netzwerk den Zugriff auf das Internet
+einschränkt (z.B. Schulen für das Schul-WLAN). Um Sharezone korrekt zu nutzen,
+müssen folgende Domains und Ports in der Firewall freigeschaltet werden.
+
```
Jonas-Sander, line 6: Maybe we could add after that, that the list is updated on an best effort basis and might not always be completely up-to-date.  
What do you think?

Jonas-Sander (approved):
LGTM, see comments

docs/more/firewall.mdx
nilsreichardt: I have moved the bottom text to the top and wrapped it around an info box, see https://docs.page/sharezoneapp/sharezone-app~667/more/firewall

Jonas-Sander: Best-effort doesn't mean "großmöglichster Mühe", more like "nach bestem Bemühen". Basically we try to keep it up-to-date if we can but it might fall under the table if we there is more important stuff that keeps us busy.

nilsreichardt: I changed it ✅

PR #685, opened 2023-07-16
Add "build android" command to Sharezone CLI

nilsreichardt:
Adds the `sz build android` command to the Sharezone CLI. It's part of a `sz deploy` command that should come in the future.

Running `sz build android --help`

```
NilsMBP13M14:sz_repo_cli nils$ fvm dart run sz_repo_cli build android --help
Build the Sharezone Android app in release mode. Codemagic CLI tools must be installed.

Usage: pub global run sz_repo_cli build android [arguments]
-h, --help            Print this usage information.
-s, --stage           [stable (default), alpha]
    --output-type     The type of output type, either "appbundle" or "apk".
                      [appbundle (default), apk]
    --build-number    An identifier used as an internal version number.
                      Each build must have a unique identifier to differentiate it from previous builds.
                      It is used to determine whether one build is more recent than
                      another, with higher numbers indicating more recent build.
                      When none is specified, the value from pubspec.yaml is used.
    --flavor          The flavor to build for. At the moment only "prod" is supported.
                      [prod (default), dev]

Run "pub global run sz_repo_cli help" to see global options.
```

Running `sz build android`

```
NilsMBP13M14:sz_repo_cli nils$ fvm dart run sz_repo_cli build android
Starting fvm flutter build appbundle --target lib/main_prod.dart --flavor prod --release --dart-define DEVELOPMENT_STAGE=STABLE...
Waiting for exit code...

Building without sound null safety ⚠️
Dart 3 will only support sound null safety, see https://dart.dev/null-safety

Running Gradle task 'bundleProdRelease'...                      
Removed unused resources: Binary resource data reduced from 609KB to 550KB: Removed 9%
Running Gradle task 'bundleProdRelease'...                         30.2s
✓ Built build/app/outputs/bundle/prodRelease/app-prod-release.aab (43.5MB).
Got exit code 0...
Build finished 🎉 
```

### DIFF
--- a/tools/sz_repo_cli/lib/src/commands/src/build_android_command.dart
+++ b/tools/sz_repo_cli/lib/src/commands/src/build_android_command.dart
@@ -59,18 +59,12 @@ When none is specified, the value from pubspec.yaml is used.''',
         help:
             'The flavor to build for. At the moment only "prod" is supported.',
         defaultsTo: 'prod',
-      )
-      ..addOption(
-        googleApplicationCredentialsOptionName,
-        help:
-            'Google Cloud service account credentials with `JSON` key type to access Google Play Developer API. If not given, the value will be checked from the environment variable `GCLOUD_SERVICE_ACCOUNT_CREDENTIALS`.',
       );
   }
 
   static const releaseStageOptionName = 'stage';
   static const flavorOptionName = 'flavor';
   static const buildNumberOptionName = 'build-number';
-  static const googleApplicationCredentialsOptionName = 'credentials';
   static const outputTypeName = 'output-type';
 
   @override
--- a/tools/sz_repo_cli/lib/src/commands/src/build_android_command.dart
+++ b/tools/sz_repo_cli/lib/src/commands/src/build_android_command.dart
@@ -82,7 +82,7 @@ When none is specified, the value from pubspec.yaml is used.''',
     isVerbose = true;
 
     await _buildApp();
-    stdout.writeln('Build finished 🎉 ');
+    print('Build finished 🎉 ');
   }
 
   Future<void> _buildApp() async {
--- a/tools/sz_repo_cli/lib/src/commands/src/build_android_command.dart
+++ b/tools/sz_repo_cli/lib/src/commands/src/build_android_command.dart
@@ -6,8 +6,6 @@
 //
 // SPDX-License-Identifier: EUPL-1.2
 
-import 'dart:io';
-
 import 'package:args/command_runner.dart';
 import 'package:sz_repo_cli/src/common/common.dart';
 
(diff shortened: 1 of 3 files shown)


tools/sz_repo_cli/lib/src/commands/src/build_android_command.dart
```diff
@@ -0,0 +1,116 @@
...
+  @override
+  String get name => 'android';
+
+  @override
+  Future<void> run() async {
+    // Its less work to just print everything right now instead of selectively
+    // print and add custom print statements for non-verboes output.
+    // One might add non-verbose output in the future but right now this is
+    // easier.
+    isVerbose = true;
```
nilsreichardt, line 83: Copied from:

https://github.com/SharezoneApp/sharezone-app/blob/c5c3ad77149f7ff580385c805c07771e69f499de/tools/sz_repo_cli/lib/src/commands/src/deploy_web_app_command.dart#L74-L78

tools/sz_repo_cli/lib/src/commands/src/build_android_command.dart
```diff
@@ -0,0 +1,116 @@
...
+another, with higher numbers indicating more recent build.
+When none is specified, the value from pubspec.yaml is used.''',
+      )
+      ..addOption(
+        flavorOptionName,
+        allowed: _androidFlavors,
+        help:
+            'The flavor to build for. At the moment only "prod" is supported.',
+        defaultsTo: 'prod',
+      );
```
Jonas-Sander, line 62: Why is only prod supported right now? (Please document)  
And if only prod is supported right now I would probably throw an Exception in the `run` method if some other flavor is passed.

(I ran `build appbundle --flavor dev` and it worked on my machine, is this just outdated?)

tools/sz_repo_cli/lib/src/commands/src/build_android_command.dart
```diff
@@ -0,0 +1,116 @@
...
+
+  @override
+  String get name => 'android';
+
+  @override
+  Future<void> run() async {
+    // Its less work to just print everything right now instead of selectively
+    // print and add custom print statements for non-verboes output.
+    // One might add non-verbose output in the future but right now this is
+    // easier.
```
Jonas-Sander, line 82:
```suggestion
    // Is used so that runProcess commands print the command that was run.
    // Right now this can't be done via an argument. 
    // This workaround should be addressed in the future.
```

I think this makes it clearer (I know its still the same in the other place but at least its better for this PR right now)

tools/sz_repo_cli/lib/src/commands/src/build_android_command.dart
```diff
@@ -0,0 +1,116 @@
...
+    isVerbose = true;
+
+    await _buildApp();
+    stdout.writeln('Build finished 🎉 ');
```
Jonas-Sander, line 86:
```suggestion
    print('Build finished 🎉 ');
```

We use `print` everywhere else so I would stick to it for now.

Jonas-Sander (approved):
See comments above, otherwise LGTM!! :)

tools/sz_repo_cli/lib/src/commands/src/build_android_command.dart
nilsreichardt: How I forgot to remove "At the moment only "prod" is supported."

PR #695, opened 2023-07-20
Implementing `sz deploy ios` for Automating iOS App Deployment

nilsreichardt:
## Description

This pull request introduces a significant enhancement to our deployment pipeline, specifically for the iOS app. It automates various stages of the deployment process, including the retrieval of the next build number, app building, and publication to the App Store or TestFlight.

Key changes include:

1. **Deployment Stages and Flavors**: The code defines different stages (`stable`, `alpha`) and flavors (`prod`) for the iOS app deployment. This allows for flexibility in managing various versions of the app.

2. **Command Line Options**: Several command-line options have been added to provide necessary parameters for the deployment process. These include API keys, issuer ID, private key, release notes, and the build flavor. 

3. **Build Number Automation**: The function `_getNextBuildNumber` is implemented to fetch the latest build number from Google Play and increment it for the next build.

4. **App Building and Publishing**: Functions for building the app (`_buildApp`) and publishing it (`_publish`) have been implemented. They use the Codemagic CLI tools and require the build number as an argument.

5. **Environment Variables**: Sensitive information, such as API keys, are fetched from environment variables, keeping them secure and out of the codebase.

This new implementation greatly simplifies the iOS deployment process, making it more streamlined and less prone to manual errors.

## Testing

* Tested the command manually by deploying a new version to the alpha group.

## Usage

`sz deploy ios --help`

```
Nils-MBP-13-M1-4:sz_repo_cli nils$ fvm dart run sz_repo_cli deploy ios --help
Deploys the Sharezone iOS app to the App Store or TestFlight. Automatically bumps the build number and builds the app. Codemagic CLI tools are required.

Usage: pub global run sz_repo_cli deploy ios [arguments]
-h, --help           Print this usage information.
-s, --stage          The stage to deploy to. The "stable" stage is used for App Store releases, the "alpha" stage is used for TestFlight releases.
                     [stable (default), alpha]
    --key-id         The App Store Connect API Key ID used to authenticate. This can be found in the App Store Connect Developer Portal. Learn more at https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api. If the parameter is not provided, the value of the APP_STORE_CONNECT_KEY_IDENTIFIER environment variable will be used. If no value is set, the deployment will fail. Example value: 1234567890
    --issuer-id      The App Store Connect API Key Issuer ID used to authenticate. This can be found in the App Store Connect Developer Portal. Learn more at https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api. If the parameter is not provided, the value of the APP_STORE_CONNECT_ISSUER_ID environment variable will be used. If no value is set, the deployment will fail. Example value: 00000000-0000-0000-0000-000000000000
    --private-key    The App Store Connect API private key used for JWT authentication to communicate with Apple services. This can be found in the App Store Connect Developer Portal. Learn more at https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api. If not provided, the key will be searched from the following directories in sequence for a private key file with the name AuthKey_<key_identifier>.p8: private_keys, ~/private_keys, ~/.private_keys, ~/.appstoreconnect/private_keys, where <key_identifier> is the value of --key-id. If not given, the value will be checked from the environment variable APP_STORE_CONNECT_PRIVATE_KEY. If no value is set, the deployment will fail.
    --whats-new      Release notes either for TestFlight or App Store review submission. Describe what's new in this version of your app, such as new features, improvements, and bug fixes. The string should not exceed 4000 characters. Example usage: --whats-new 'Bug fixes and performance improvements.'
    --flavor         The flavor to build for. Only the "prod" flavor is supported.
                     [prod (default)]

Run "pub global run sz_repo_cli help" to see global options.
```

### DIFF
--- a/tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
+++ b/tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
@@ -71,7 +71,7 @@ class DeployIosCommand extends Command {
       ..addOption(
         privateKeyOptionName,
         help:
-            'The App Store Connect API private key used for JWT authentication to communicate with Apple services. This can be found in the App Store Connect Developer Portal. Learn more at https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api. If not given, the value will be checked from the environment variable APP_STORE_CONNECT_PRIVATE_KEY. If not provided, the key will be searched from the following directories in sequence for a private key file with the name AuthKey_<key_identifier>.p8: private_keys, ~/private_keys, ~/.private_keys, ~/.appstoreconnect/private_keys, where <key_identifier> is the value of --key-id. If no value is set, the deployment will fail.',
+            'The App Store Connect API private key used for JWT authentication to communicate with Apple services. This can be found in the App Store Connect Developer Portal. Learn more at https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api. If not provided, the value will be checked from the environment variable APP_STORE_CONNECT_PRIVATE_KEY. If not given, the key will be searched from the following directories in sequence for a private key file with the name AuthKey_<key_identifier>.p8: private_keys, ~/private_keys, ~/.private_keys, ~/.appstoreconnect/private_keys, where <key_identifier> is the value of --key-id. If no value is set, the deployment will fail.',
       )
       ..addOption(
         whatsNewOptionName,
--- a/tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
+++ b/tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
@@ -1,0 +1,225 @@
+// Copyright (c) 2022 Sharezone UG (haftungsbeschränkt)
+// Licensed under the EUPL-1.2-or-later.
+//
+// You may obtain a copy of the Licence at:
+// https://joinup.ec.europa.eu/software/page/eupl
+//
+// SPDX-License-Identifier: EUPL-1.2
+
+import 'dart:io';
+
+import 'package:args/command_runner.dart';
+import 'package:sz_repo_cli/src/common/common.dart';
+
+final _iosStages = [
+  'stable',
+  'alpha',
+];
+
+/// The different flavors of the iOS app that support deployment.
+final _iosFlavors = [
+  'prod',
+];
+
+class DeployIosCommand extends Command {
+  final SharezoneRepo _repo;
+
+  DeployIosCommand(this._repo) {
+    argParser
+      ..addOption(
+        releaseStageOptionName,
+        abbr: 's',
+        allowed: _iosStages,
+        defaultsTo: 'stable',
+      )
+      ..addOption(
+        keyIdOptionName,
+        help:
+            'The App Store Connect API Key ID used to authenticate. This can be found in the App Store Connect Developer Portal. Learn more at https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api. If the parameter is not provided, the value of the APP_STORE_CONNECT_KEY_IDENTIFIER environment variable will be used. If no value is set, the deployment will fail.\nExample value: 1234567890',
+      )
+      ..addOption(
+        issuerIdOptionName,
+        help:
+            'The App Store Connect API Key Issuer ID used to authenticate. This can be found in the App Store Connect Developer Portal. Learn more at https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api. If the parameter is not provided, the value of the APP_STORE_CONNECT_ISSUER_ID environment variable will be used. If no value is set, the deployment will fail.\nExample value: 00000000-0000-0000-0000-000000000000',
+      )
+      ..addOption(
+        privateKeyOptionName,
+        help:
+            'The App Store Connect API private key used for JWT authentication to communicate with Apple services. This can be found in the App Store Connect Developer Portal. Learn more at https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api. If not provided, the key will be searched from the following directories in sequence for a private key file with the name AuthKey_<key_identifier>.p8: private_keys, ~/private_keys, ~/.private_keys, ~/.appstoreconnect/private_keys, where <key_identifier> is the value of --key-id. If not given, the value will be checked from the environment variable APP_STORE_CONNECT_PRIVATE_KEY. If no value is set, the deployment will fail.',
+      )
+      ..addOption(
+        whatsNewOptionName,
+        help:
+            "Release notes either for TestFlight or App Store review submission. Describe what's new in this version of your app, such as new features, improvements, and bug fixes. The string should not exceed 4000 characters. Example usage: --whats-new 'Bug fixes and performance improvements.'",
+      )
+      ..addOption(
+        flavorOptionName,
+        allowed: _iosFlavors,
+        help: 'The flavor to build for. Only the "prod" flavor is supported.',
+        defaultsTo: 'prod',
+      );
+  }
+
+  static const privateKeyOptionName = 'private-key';
+  static const keyIdOptionName = 'key-id';
+  static const issuerIdOptionName = 'issuer-id';
+  static const releaseStageOptionName = 'stage';
+  static const flavorOptionName = 'flavor';
+  static const whatsNewOptionName = 'whats-new';
+
+  @override
+  String get description =>
+      'Deploys the Sharezone iOS app to the App Store or TestFlight. Automatically bumps the build number and builds the app. Codemagic CLI Tools are required.';
+
+  @override
+  String get name => 'ios';
+
+  @override
+  Future<void> run() async {
+    _throwIfFlavorIsNotSupportForDeployment();
+
+    // Is used so that runProcess commands print the command that was run. Right
+    // now this can't be done via an argument.
+    //
+    // This workaround should be addressed in the future.
+    isVerbose = true;
+
+    final buildNumber = await _getNextBuildNumber();
+    await _buildApp(buildNumber: buildNumber);
+    await _publish();
+
+    print('Deployment finished 🎉 ');
+  }
+
+  void _throwIfFlavorIsNotSupportForDeployment() {
+    final flavor = argResults![flavorOptionName] as String;
+    if (flavor != 'prod') {
+      throw Exception(
+        'Only the "prod" flavor is supported for iOS deployment.',
+      );
+    }
+  }
+
+  Future<void> _publish() async {
+    final whatsNew = argResults![whatsNewOptionName] as String?;
+    final stage = argResults![releaseStageOptionName] as String;
+    final issuerId = argResults![issuerIdOptionName] as String? ??
+        Platform.environment['APP_STORE_CONNECT_ISSUER_ID'];
+    final keyIdentifier = argResults![keyIdOptionName] as String? ??
+        Platform.environment['APP_STORE_CONNECT_KEY_IDENTIFIER'];
+    final privateKey = argResults![privateKeyOptionName] as String? ??
+        Platform.environment['APP_STORE_CONNECT_PRIVATE_KEY'];
+
+    final isStable = stage == 'stable';
+    await runProcess(
+      'app-store-connect',
+      [
+        'publish',
+        '--path',
+        'build/ios/ipa/*.ipa',
+        '--release-type',
+        'AFTER_APPROVAL',
+        if (whatsNew != null) ...[
+          '--whats-new',
+          whatsNew,
+        ],
+        if (!isStable) ...[
+          '--beta-group',
+          stage,
+          '--testflight',
+        ],
+        if (issuerId != null) ...[
+          '--issuer-id',
+          issuerId,
+        ],
+        if (keyIdentifier != null) ...[
+          '--key-id',
+          keyIdentifier,
+        ],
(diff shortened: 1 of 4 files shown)


Jonas-Sander:
Gonna review it completely tomorrow (sorry! ;p), but one "bigger" thing I'm not sure about right now is if we should always build the app in the "deploy" command. For me that is a bit unexpected. 

I guess there could be the default case that no prod app was built, so it will be build when running the deploy command. In this case if there is already a prod app built then it would be used automatically? 
Or maybe, since this could be complicating it unnecessarily, lets just add a `--build-app` flag for now? Or we just leave it like that since we both know that it builds the app and I'm just overthinking. Although what about if the deployment fails because some other flag was wrong and now we have to build the whole app again since we have no choice here? I think that an extra `--build-app` flag might be the best compromise for now. What do you think?

tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
```diff
@@ -0,0 +1,254 @@
...
+  String get description =>
+      'Deploys the Sharezone iOS app to the App Store or TestFlight. Automatically bumps the build number and builds the app. Codemagic CLI tools are required.';
+
+  @override
+  String get name => 'ios';
+
+  @override
+  Future<void> run() async {
+    _throwIfFlavorIsNotSupportForDeployment();
+
```
Jonas-Sander, line 107: Nit: Might it be more user friendly to check if the codemagic CLI tools are installed right here and if not to throw an error with a description on how to install them? I guess this might be done with `which app-store-connect` (or `which -s app-store-connect`, not sure).

tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
```diff
@@ -0,0 +1,254 @@
...
+    // Is used so that runProcess commands print the command that was run. Right
+    // now this can't be done via an argument.
+    //
+    // This workaround should be addressed in the future.
+    isVerbose = true;
+
+    final buildNumber = await _getNextBuildNumber();
+    await _buildApp(buildNumber: buildNumber);
+    await _publish();
+
```
Jonas-Sander, line 117: Nice, very apprehensible code!

Nit: order the method implementations below in the same order (so that after `_throwIfFlavorIsNotSupportForDeployment` below comes `_getNextBuildNumber`, then `_buildApp`, then `_publish`). This helps to understand the code more quickly becaues you're going with the flow of the code when scrolling down in the IDE (looking at each method implementation in the order it will be executed)

tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
```diff
@@ -0,0 +1,254 @@
...
+    final isStable = stage == 'stable';
+    await runProcess(
+      'app-store-connect',
+      [
+        'publish',
+        '--path',
+        'build/ios/ipa/*.ipa',
+        '--release-type',
+        'AFTER_APPROVAL',
```
Jonas-Sander, line 148:
```suggestion
        // The app version will be automatically released right after
        // it has been approved by App Review.
        '--release-type',
        'AFTER_APPROVAL',
```

I e.g. thought that this would mean that the version is only published after **we** approved it manually in the app store connect website.

(Note: Comment might be wrongly formatted)

tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
```diff
@@ -0,0 +1,254 @@
...
+            '--private-key',
+            privateKey,
+          ],
+        ],
+        // Using the app location as working direcorty because the default
+        // location for the App Store Connect private key is
+        // app/private_keys/AuthKey_{keyIdentifier}.p8.
+        workingDirectory: _repo.sharezoneFlutterApp.location.path,
+      );
+      return int.parse(result.stdout);
```
Jonas-Sander, line 222: Since this is `runProcess` instead of `runProcessSucessfullyOrThrow` you haven't actually handled an error case.  
I think you should replace it with `runProcessSucessfullyOrThrow`. For this you will probably have to change the `runProcessSucessfullyOrThrow` to return a `ProcessResult`.

tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
```diff
@@ -0,0 +1,254 @@
...
+    final stage = argResults![releaseStageOptionName] as String;
+    final issuerId = argResults![issuerIdOptionName] as String? ??
+        Platform.environment['APP_STORE_CONNECT_ISSUER_ID'];
+    final keyIdentifier = argResults![keyIdOptionName] as String? ??
+        Platform.environment['APP_STORE_CONNECT_KEY_IDENTIFIER'];
+    final privateKey = argResults![privateKeyOptionName] as String? ??
+        Platform.environment['APP_STORE_CONNECT_PRIVATE_KEY'];
+
+    final isStable = stage == 'stable';
+    await runProcess(
```
Jonas-Sander, line 141:
```suggestion
    await runProcessSucessfullyOrThrow(
```

Jonas-Sander (changes requested):

tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
```diff
@@ -0,0 +1,254 @@
...
+      )
+      ..addOption(
+        privateKeyOptionName,
+        help:
+            'The App Store Connect API private key used for JWT authentication to communicate with Apple services. This can be found in the App Store Connect Developer Portal. Learn more at https://developer.apple.com/documentation/appstoreconnectapi/creating_api_keys_for_app_store_connect_api. If not provided, the key will be searched from the following directories in sequence for a private key file with the name AuthKey_<key_identifier>.p8: private_keys, ~/private_keys, ~/.private_keys, ~/.appstoreconnect/private_keys, where <key_identifier> is the value of --key-id. If not given, the value will be checked from the environment variable APP_STORE_CONNECT_PRIVATE_KEY. If no value is set, the deployment will fail.',
```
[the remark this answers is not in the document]
nilsreichardt: You don't need to communicate with App Store Connect to be able to build the app (`flutter build ipa`). Communication with App Store Connect is only required when asking App Store Connect for the latest build number (`app-store-connect get-latest-build-number`) and publishing the build to App Store Connect (`app-store-connect publish`). I can add this to help to clarify this.

The order of the sentences is correct, they are copied from the `app-store-connect publish` command: https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/publish.md#--private-keyprivate_key

1. Try to use the provided argument
2. Try to use the key from the file
3. Try to use from `APP_STORE_CONNECT_PRIVATE_KEY`

tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
nilsreichardt: Cool, didn't know "which -s" existed.

tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
```diff
@@ -0,0 +1,254 @@
...
+    argParser
+      ..addOption(
+        releaseStageOptionName,
+        abbr: 's',
+        allowed: _iosStages,
+        help:
+            'The stage to deploy to. The "stable" stage is used for App Store '
+            'releases, the "alpha" stage is used for TestFlight releases.',
+        defaultsTo: 'stable',
+      )
```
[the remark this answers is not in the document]
nilsreichardt: > Random note: For the deploy web command the have the stages [alpha, beta, prod] instead of [alpha, beta, stable]. I guess these should be the same everywhere?

For now, we don't have an iOS beta track. Web is the only platform with a beta track (which btw we don't use xD).

> Also can you describe in the help what the stages do exactly and how they are used? From what I gathered "--stage alpha" will need a matching "beta group" in TestFlight called "alpha". Additonally the stage is used in the build process, what will it change there? Is there a change?

Do you think it's a good idea to describe what the stage parameter changes in the build process / in the app at the help string of the iOS deploy command? I'll also create a PR for `sz deploy android` and `sz deploy macos` command. Changing the behavior how `--dart-define=DEVELOPEMTENT_STAGE` affects the app behavior would require to update the documentation at 5 difference locations (inside the app, at `sz deploy ios`, `sz deploy android`, `sz deploy web` and `sz deploy macos`).

My current draft of the `help` text is:

The stage to deploy to. The "stable" stage is used for App Store releases, the "alpha" stage is used for TestFlight releases (publishing to a TestFligh group that matches "alpha" as a group name). Additionally, the stage value is passed to the build command.

nilsreichardt:
> Gonna review it completely tomorrow (sorry! ;p), but one "bigger" thing I'm not sure about right now is if we should always build the app in the "deploy" command. For me that is a bit unexpected.
> 
> I guess there could be the default case that no prod app was built, so it will be build when running the deploy command. In this case if there is already a prod app built then it would be used automatically for deployment?
> 
> Or maybe, since this could be complicating it unnecessarily, lets just add a `--build-app` flag for now? Or we just leave it to always build (current logic). Although what about if the deployment fails because some other flag was wrong and now we have to build the whole app again since we have no choice here? I think that an extra `--build-app` flag might be the best compromise for now. What do you think?
> 
> Ah and for the first approach I don't know how easy it would be to check if an already built app has the `prod` flavor. An easier approach could also be to just assume it is as implicit knowledge from our side (but which could lead to accidentally uploading a non-prod version). If its easy to check that its `prod` flavor then I would maybe favor the first approach, otherwise the `--build-app` (or whatever) approach. Or we could have the first approach with an additional `--force-build-app` flag.... Ah idk, i'm too tired right now:(

My main intention for the `sz deploy` commands is to add a `stable.yml` CD file to our `.github/workflows/` folder to be able to press one button to publish a new stable version.

Technically, it's just copying the [`alpha.yml`](https://github.com/SharezoneApp/sharezone-app/blob/main/.github/workflows/alpha.yml) file (for Android we would need to use `fastlane` instead of Firebase Distribution) and make a few changes (publish to App Store instead to TestFlight for iOS). However, to reduce code duplication (long-term goal is also to have a `beta.yml`), I started to create the `sz deploy` commands.

For this reason, the `sz deploy` commands are designed to be executed in our CD pipeline. Because when we have a working CD pipeline, there should be no reason to publish a new version with your local machine.

With this in mind I think having things like `--build-app`, etc. are a bit overkill because that are things we would (ideally) never use and it would increase the complexity of the code for future scenarios that are not planned. I think the approach "If we actually need these things, we can always add them later" is here the better way.

tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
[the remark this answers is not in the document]
nilsreichardt: You're right 🙏 Fixed it 👍

tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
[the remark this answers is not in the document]
Jonas-Sander: It was just confusing for me because for me just reading the argument description I thought it was **only** for TestFlight which confused me because I saw it was also passed to the build command (I forgot all about stages :D).

So I guess one thing I don't like is that the stage name and the TestFlight beta group names are coupled. In my opinion they should be decoupled so that we could change the stage names and not break anything for TestFlight. Decoupling it would have also helped me to understand that the stage string is not primarily for TestFlight, but our definition of deployment stages. They could be decoupled like in the `deploy web-app` code (just a simple map should be enough I guess).

> Do you think it's a good idea to describe what the stage parameter changes in the build process / in the app at the help string of the iOS deploy command? I'll also create a PR for sz deploy android and sz deploy macos command. Changing the behavior how --dart-define=DEVELOPEMTENT_STAGE affects the app behavior would require to update the documentation at 5 difference locations (inside the app, at sz deploy ios, sz deploy android, sz deploy web and sz deploy macos).

Yeah you're right, a detailed description would be bad. 

I like your description. If we decoupled the TestFlight stage name then we could leave the stuff about the beta group out in my opinion.

What about this?
> The deployment stage to deploy to. The "stable" stage is used for App Store releases, the "alpha" stage is used for TestFlight releases. The value will be forwarded to the "sz build" command (if a fresh build is necessary).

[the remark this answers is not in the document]
nilsreichardt: I'm using now the following the `help` value for `--stage`:

"The deployment stage to deploy to. The "stable" stage is used for App Store releases, the "alpha" stage is used for TestFlight releases. The value will be forwarded to the "sz build" command "

I also decoupled the stage and the TestFligh Group names 👍

tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
```diff
@@ -0,0 +1,304 @@
...
+
+/// The track for TestFlight.
+///
+/// https://appstoreconnect.apple.com/apps/1434868489/testflight
+class TestFlightTrack extends AppleTrack {
+  /// The name of the TestFlight group.
+  final String groupName;
+
+  const TestFlightTrack(this.groupName);
+}
```
nilsreichardt: This will be reused for macOS. Not sure where the file should be located. In `lib/src/common/src`?

Jonas-Sander: Yes, sounds good 👍

Jonas-Sander (approved):
LGTM 👍

PR #697, opened 2023-07-21
Optimize App Startup by Background Loading of Firebase Remote Config

### LINKED ISSUE #696: Sharezone takes a long time to start with a poor internet connection
https://github.com/SharezoneApp/sharezone-app/assets/24459435/c90b80f4-7b20-435b-9516-daee6e5b2f87

Used as EDGE network type. Stopped: ~ 11 seconds to start (from second 2 to second 13).

nilsreichardt:
## Description

In this pull request, we made significant changes to the way our application fetches and activates the Firebase Remote Config data to improve app start-up times.

Previously, our app fetched Firebase Remote Config data during its start-up sequence. This fetch-and-activate approach was a major contributor to the 11 seconds startup time our app was experiencing.

We have now moved the Firebase Remote Config fetch process to execute in the background after the application has started. This ensures that the app is not blocked waiting for these config values during startup. The fetched config values are activated on the next app launch. 

This change has led to a considerable improvement in app startup times. Initial tests show startup times have dropped to just 2 seconds, down from 11 seconds - a more than 5x improvement.

### Disadvantages in this strategy

The disadvantage in this strategy is that the values activated on the next launch. When there is a critical change, the user doesn't receive the values immediately. However, since we started Sharezone, 5 years ago, we never had this case. Therefore, I think it's overkill to trade bad app start times against the possibility to update critical values maybe someday.

## Benchmarks

| Name | Time |
|-|-|
| Before the changes | 11 seconds |
| After the changes | 2 seconds |

The app start is now 5,5x faster than before when using an EDGE connection.

### Demo

(In the benchmarks table, I removed the time I needed to open the emulator and click the "stop" button)

#### Before 

https://github.com/SharezoneApp/sharezone-app/assets/24459435/f2c512ac-0a4c-45df-8f22-afff05611d87

#### After

https://github.com/SharezoneApp/sharezone-app/assets/24459435/de172d53-1147-43b5-b1df-f228ce5b51ec

#### Using the values for the next start

https://github.com/SharezoneApp/sharezone-app/assets/24459435/198add26-2da3-4fb2-94ac-59e8f3defe0e

## Related Ticktes

Fixes #696

### DIFF
--- a/lib/remote_configuration/lib/src/implementation/firebase_remote_configuration.dart
+++ b/lib/remote_configuration/lib/src/implementation/firebase_remote_configuration.dart
@@ -6,6 +6,7 @@
 //
 // SPDX-License-Identifier: EUPL-1.2
 
+import 'dart:async';
 import 'dart:developer';
 
 import 'package:firebase_remote_config/firebase_remote_config.dart';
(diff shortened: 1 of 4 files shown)


lib/remote_configuration/lib/src/implementation/firebase_remote_configuration.dart
```diff
@@ -29,15 +30,33 @@ class FirebaseRemoteConfiguration extends RemoteConfiguration {
...
+      );
+      // Activate the fetched remote config from the last fetch.
+      await _remoteConfig.activate();
+
+      // Fetch remote config in the background. They will be available on the
+      // next app start.
+      unawaited(_remoteConfig.fetch());
     } catch (e) {
       log("Error fetch remote config: $e");
     }
```
Jonas-Sander, line 62: Couldn't we just leave the initialize method with the old logic and add the `initializeAndFetchInBackground` method that just calls `unawaited(initialize)`?

Jonas-Sander: No strong need but I also don't see why we would have to remove initialize entirely... (If we leave it should we call it `intializeAndFetch` as well then?)

Jonas-Sander (approved):
LGTM, see comments

lib/remote_configuration/lib/src/implementation/firebase_remote_configuration.dart
nilsreichardt: I refactored the code to a new style. I removed all the activate and fetch logic from the `initialize` method (which makes for more sense that the `initialize` method only initializes), added a `fetch` and `activate` method to the package so that we can use the "Load new values for next startup" strategy the `plugin_initializations` file.

Jonas-Sander (approved):
LGTM, I like the new approach.   

Couldn't initialize be synchronous now? `void initialize(...)`

nilsreichardt:
> Couldn't initialize be synchronous now? void initialize(...)

Yes, good catch!

PR #699, opened 2023-07-30
Add CD workflow to deploy new iOS stable versions

nilsreichardt:
This pull request introduces a new GitHub Actions workflow for deploying new stable versions of our application.

The workflow, named 'stable', ensures that our iOS builds are built sequentially to maintain unique build numbers, a requirement of iOS builds. The concurrency group, 'app-release', enforces that only one version is built per branch at a time.

A manual trigger has been implemented via the workflow_dispatch event. This requires the input of a changelog for the iOS release. This changelog is subsequently used as the release notes in the App Store release.

### DIFF
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -37,9 +37,6 @@ jobs:
     if: github.event.inputs.ios-changelog != ''
     runs-on: macos-13
     timeout-minutes: 120
-    defaults:
-      run:
-        working-directory: app
     steps:
       - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
 
@@ -47,6 +44,7 @@ jobs:
         run: pip3 install codemagic-cli-tools==0.39.1
 
       - name: Setup signing
+        working-directory: app
         env:
           # The following secrets are used by the Codemagic CLI tool. It's important
           # to use the same names as the CLI tool expects.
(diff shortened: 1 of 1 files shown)


.github/workflows/stable.yml
```diff
@@ -0,0 +1,91 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      ios-changelog:
+        type: string
+        description: |
+          The changelog for the iOS release. This will be used as the release
+          notes in the App Store release.
+        required: true
```
nilsreichardt, line 26: Or should we add an option that you can select which platforms should get a new stable release? Imagine that we fix a web only bug and we want only deploy web.

Jonas-Sander: Yes, that sounds like a nice idea. Good thinking! :)

Jonas-Sander (approved):
Nice, LGTM 👍

.github/workflows/stable.yml
nilsreichardt: I have changed the description to:

> iOS changelog: Used for the App Store release notes. If the changelog is not provided, the deployment for iOS will be skipped.

Assuming that if there is no changelog, nothing has changed for iOS and no new deployment is needed.

nilsreichardt:
Status: I need also apply the fixes from #709, #708 and #707

PR #630, opened 2023-04-30
Introduce alpha program for our macOS app

nilsreichardt:
Adds the alpha program for macOS. Currently, the implementation does not use the `sz deploy macos` command because I need to adjust the command (the command also needs to do the signing). This will be done in a future PR.

PR #715, opened 2023-08-06
Re-configure macOS

nilsreichardt:
* Deleted the `app/macos` folder in the hope that it could fix #630
* Created a new `app/macos` folder with `fvm flutter create . --platforms=macos`
* Set up all the necessary things:
  * Set app name to "Sharezone"
  * App logo
  * Firebase setup
  * Signing
* Tested manually


app/macos/Podfile
```diff
@@ -38,8 +38,5 @@ end
 post_install do |installer|
   installer.pods_project.targets.each do |target|
     flutter_additional_macos_build_settings(target)
-    target.build_configurations.each do |bc|
-      bc.build_settings['MACOSX_DEPLOYMENT_TARGET'] = '10.14'
-    end
```
Jonas-Sander, line 43: This is not important?

Jonas-Sander (approved):

app/macos/Podfile
nilsreichardt: We needed this for M1 stuff back in the days. Also worked without it, so I would leave it for now. If a similar error occurs, we can add it again.

PR #734, opened 2023-08-09
Add `sharezone_lints` to `abgabe_http_api`

nilsreichardt:
Part of #37


Jonas-Sander (approved):
Since this is an auto-generated library, I don't really think that this makes sense. But since you already changed it, its okay to merge.

PR #738, opened 2023-08-16
Upgrade bloc library in `hausaufgabenheft_logik` package.

### LINKED ISSUE #729: Upgrade `bloc` library in `hausaufgabenheft_logik`  package.
Blocks #191

Jonas-Sander:
Upgrade the `bloc` package from `^4.0.0` to `^8.1.2`.    
This enables us to make the `hausaufgabenheft_logik` package null-safe afterwards (#191). 

Note: Before there was an abstract "interface" for some blocs in our code, which I removed because of class types issues (Bloc was `Stream` before but isn't after upgrading) and also because it's unnecessary. Thus I also renamed some bloc files which git seems to interpret as deleting old bloc files and creating new ones.

Closes #729


nilsreichardt (commented):
LGTM

nilsreichardt (approved):
LGTM

PR #740, opened 2023-08-17
Migrate `group_domain_implementation` to null safety

### LINKED ISSUE #193: Migrate the `group_domain_implementation` package to null safety
Blocked by:
* #192 
* #187 
* #182

nilsreichardt:
Closes #193


lib/group_domain_implementation/lib/src/accessors/firestore_course_member_accessor.dart
```diff
@@ -110,13 +123,13 @@ class FirestoreCourseMemberAccessor extends CourseMemberAccessor {
         if (docSnapshot.exists) {
           joinedUsersData = MemberData.create(
             id: docSnapshot.id,
-            role: docSnapshot.data()['powerLevel'] == 'owner'
+            role: docSnapshot.data()!['powerLevel'] == 'owner'
```
nilsreichardt, line 126: Should we just assume that these attributes are present?

Jonas-Sander: Yes, since we're only assuming that there is data at all, which should always be true for a course.

Jonas-Sander (approved):

PR #746, opened 2023-08-17
Format GitHub Actions files with `sz format` command

nilsreichardt:
* Adds `npx prettier ".github/workflows/" --check` as source of truth command
* Add `throwIfFvmIsNotInstalled` method
* Add `runSourceOfTruthCommand` method to run source of commands
* Use `runSourceOfTruthCommand` method for license header


tools/sz_repo_cli/lib/src/commands/src/format_command.dart
```diff
@@ -24,7 +26,46 @@ class FormatCommand extends ConcurrentCommand {
...
+    await _formatActionFiles(repo: repo);
+  }
+
+  @override
+  Future<void> runTaskForPackage(Package package) async =>
+      await formatCode(package);
+
+  Future<void> _throwIfPrettierIsNotInstalled() async {
+    // Check if "which -s app-store-connect" returns 0.
+    // If not, throw an exception.
```
Jonas-Sander, line 42: old comment

Jonas-Sander (approved):

PR #753, opened 2023-08-18
Migrate `sharezone` package to null safety

### LINKED ISSUE #213: Migrate the `sharezone` package to null safety
Blocked by:
* #211 
* #210 
* #209 
* #208
* #207
* #206
* #205
* #204
* #203
* #202
* #201
* #200
* #199
* #198
* #197
* #196
* #195
* #194
* #193
* #192
* #191
* #190
* #189
* #188
* #187
* #186
* #185
* #184
* #183
* #182
* #181

nilsreichardt:
## ToDo

* [ ] Check if there are broken user documents
* [ ] Check if there are cloud files with `name == null`

## Related Tickets

Closes #213


nilsreichardt:
In `blackboard_item.dart` we just remove the `authorReference`. In production, we don't have any document where `authorID` is missing, so we can just use `auhtorID` instead of `authorReference`.

I exported the collection and used this query in Big Query:

```sql
SELECT __key__.name
FROM `sharezone-c2bd8.null_safety_migration.blackboard`
WHERE authorID IS NULL
AND authorReference IS NOT NULL
```

Jonas-Sander (approved):
(Since this is so much change we agreed on merging this without a proper review.)

PR #761, opened 2023-08-21
Add git diff checks for integration tests

nilsreichardt:
We're already catching Flutter dependency changes in the `analyze` check. However, to catch platform-specific file changes, I added the git diff check to the integration test checks.

### DIFF
--- a/.github/workflows/integration_tests_app_ci.yml
+++ b/.github/workflows/integration_tests_app_ci.yml
@@ -154,6 +154,12 @@ jobs:
             --test ../build/app/outputs/apk/androidTest/prod/debug/app-prod-debug-androidTest.apk \
             --device model=Pixel2,version=30,locale=en,orientation=portrait \
             --timeout 3m
+    
+      # It can easily happen that a dependency changed but the .lock file is not
+      # updated. Or other cases where files are changed during a build.
+      # Therefore, fails this check if there are Git changes.
+      - name: Fail if there are Git diffs
+        run: git diff --exit-code
 
   ios-integration-test:
     needs: changes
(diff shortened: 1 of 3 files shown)


.github/workflows/integration_tests_app_ci.yml
```diff
@@ -154,6 +154,12 @@ jobs:
             --test ../build/app/outputs/apk/androidTest/prod/debug/app-prod-debug-androidTest.apk \
             --device model=Pixel2,version=30,locale=en,orientation=portrait \
             --timeout 3m
+    
+      # It can easily happen that a dependency changed but the .lock file is not
+      # updated. Or other cases where files are changed during a build.
+      # Therefore, fails this check if there are Git changes.
+      - name: Fail if there are Git diffs
+        run: git diff --exit-code --name-status
```
nilsreichardt, line 162: @Jonas-Sander FYI: I added the `--name-status` so that only the file names are printed:

```
$ git diff --exit-code --name-status
M	app/pubspec.lock
M	app/pubspec.yaml
```

nilsreichardt: I can add this in a new PR for the other checks

nilsreichardt:
Ok, this check doesn't make sense for macOS and iOS. When the checksum changes somehow and on macOS XCode files are changed because of signing the app.

.github/workflows/integration_tests_app_ci.yml
nilsreichardt: > I can add this in a new PR for the other checks

I updated it in this PR.

Jonas-Sander (approved):
Is it possible to provide the changed files as output that e.g. if you're on Windows and Mac files changed that you could at least copy them from the GitHub Action?  

Of course we need to somehow make sure that these can't be sensitive files or that only we can access the output.

nilsreichardt:
> Is it possible to provide the changed files as output that e.g. if you're on Windows and Mac files changed that you could at least copy them from the GitHub Action?

Technically, yes this is possible. However, because it has some edge cases (as you mentioned with the sensitive data) it I would wait and track how big this problem is and if it's worth implementing it. We don't have this checks for macOS and iOS for now, therefore it should only fail because of Flutter (e.g. forgot to run `flutter pub get`) or Android stuff and Flutter and Android changes should be easily fixable with local command on every machine.

Issue #762, opened 2023-08-21
Watching a video does not work


nilsreichardt:
![IMG_D7A4B869E7E9-1](https://github.com/SharezoneApp/sharezone-app/assets/24459435/4f3a4918-6b54-4231-a210-d8933410e6ae)

## Steps to reproduce

1. Open Sharezone
2. Upload a video
3. Try to open it

## Logs

```
flutter: ----------------FIREBASE CRASHLYTICS----------------
flutter: PlatformException(video_player, not implemented, null, null)
flutter:
#0      AVFoundationVideoPlayerApi.create
messages.g.dart:303
<asynchronous suspension>
#1      AVFoundationVideoPlayer.create
avfoundation_video_player.dart:65
<asynchronous suspension>
#2      VideoPlayerController.initialize
video_player.dart:362
<asynchronous suspension>
flutter: ----------------------------------------------------
flutter: ══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
flutter: The following _CastError was thrown while finalizing the widget tree:
flutter: Null check operator used on a null value
flutter:
flutter: When the exception was thrown, this was the stack:
flutter: #0      __VideoViewerState.dispose
video_viewer.dart:48
flutter: #1      StatefulElement.unmount
framework.dart:5179
flutter: #2      _InactiveElements._unmount
framework.dart:1960
flutter: #3      _InactiveElements._unmount.<anonymous closure>
framework.dart:1958
```

nilsreichardt:
Not sure what the problem there is. When loading it for the first time, it is not working. Trying it a second time, does word.

PR #766, opened 2023-08-21
Fix "Null check operator used on a null value" when when closing the video player

nilsreichardt:
Related to #762


nilsreichardt:
Not sure what the problem there is. When loading it for the first time, it is not working. Trying it a second time, does word.

PR #779, opened 2023-08-22
Fix to allow entering time values for each minute when adding an event or lesson

### LINKED ISSUE #778: Adding an event does not allow entering times when `minutes % 5 != 0`
image

nilsreichardt:
## Demo

https://github.com/SharezoneApp/sharezone-app/assets/24459435/a49ebd46-214d-4d25-b2e6-97454e6dab20

## Related tickets

Fixes #778


Jonas-Sander:
Wouldn't this make picking the time slower on mobile on iOS as now you would need to scroll through all the minutes instead of going through 5 minute intervals?

nilsreichardt:
> Wouldn't this make picking the time slower on mobile on iOS as now you would need to scroll through all the minutes instead of going through 5 minute intervals?

Yes, good catch! We can just ignore the `isFiveMinutesIntervalActive` option when we are not on iOS because for other platforms we have the interval time picker where we have the visual steps option.

Jonas-Sander (approved):

PR #782, opened 2023-08-22
Improve `UpdatePromptCard`

nilsreichardt:
* Do not show "download new version from Play Store" on the web anymore
* Add golden tests
* Use "https://sharezone.net/mobile-store" instead of "https://onelink.to"
* Write "App Store" and "Play Store" instead of "AppStore" and "PlayStore"
* Translate comment to English


app/lib/pages/settings/changelog_page.dart
```diff
@@ -168,33 +168,38 @@ class _ChangeParagraph extends StatelessWidget {
...
-    // wird nur in den Store geleitet, wenn dieser die App nicht installiert hat.
-    //
-    // Der "onelink.to" leitet immer in den Store - egal, ob die App installiert
-    // ist oder nicht.
-    return "http://onelink.to/eqkmz5";
+    // It should NOT be used for iOS & Android https://sharezone.net/ios or
+    // https://sharezone.net/android because the user will be taken straight
+    // back to the app once it is installed. The user is only redirected to the
+    // store if the app is not installed. has.
+    return "http://sharezone.net/mobile-store";
```
Jonas-Sander, line 203:
```suggestion
    // Instead of using https://sharezone.net/ios or
    // https://sharezone.net/android we use the link below
    // because it will always lead to the app store, regardless
    // of whether the app is already installed or not.
    // The iOS and Android links would just lead the user
    // back to this app (since the app is already installed)
    // which is not what we want since the user should tap 
    // "update" in the app store.
    return "http://sharezone.net/mobile-store";
```

Jonas-Sander (approved):
I personally don't know if we really should use such a harsh red color. I guess it was the same color before, but I just imagine that if we use it too often users might just start to ignore it (and other red announcement cards) in the future.
Maybe only if the app version is really old (altough in this case we could think about forcing the user to upgrade).  

Hm, I'm not sure myself.
Just a thought, we can keep it like this.

app/lib/pages/settings/changelog_page.dart
nilsreichardt: I also noticed that https://sharezone.net/android are opening the play store and not the app. Therefore, the comment is incorrect and can be deleted. This was the case https://sharezone.net/app because it uses a Firebase Dynamic Link but not for the normal https://android & ios links.

nilsreichardt:
> I personally don't know if we really should use such a harsh red color.

I agree. I changed the color to orange.

>  but I just imagine that if we use it too often users might just start to ignore it (and other red announcement cards) in the future.
Maybe only if the app version is really old (altough in this case we could think about forcing the user to upgrade).

In theory, this is correct. However, you have to that this dialog is only shown 3 days after release where most of the people should have the version. In the long term, we should use the native updater that Android provides (only applies for Android) which allows updating without leaving the app (downloads the update and automatically restarts)

The updater:

image


The grace period for showing the dialog:

https://github.com/SharezoneApp/sharezone-app/blob/4ae0a6470a0f928b3c11dbabeb55fedccf642cb7/app/lib/blocs/sharezone_bloc_providers.dart#L349

app/lib/pages/settings/changelog_page.dart
```diff
@@ -168,33 +168,38 @@ class _ChangeParagraph extends StatelessWidget {
 }
 
 class UpdatePromptCard extends StatelessWidget {
+  String _getText() {
+    if (PlatformCheck.isWeb) {
+      return "Wir haben bemerkt, dass du eine veraltete Version der App verwendest. Lade die Seite neu, um die neuste Version zu erhalten! 👍";
+    }
```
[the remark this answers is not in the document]
nilsreichardt: For me, it's not a problem anymore or I never experienced a problem with Flutter web apps I build. I think it has something to do that we are the canvaskit renderer using. In the first days of the Sharezone Web canvaskit wasn't ready enough, so we used the HTML renderer.

PR #783, opened 2023-08-22
Fix changelog showing always "please update the update"

### LINKED ISSUE #781: Changelog shows that there is a newer version even when the newest version is installed
image

nilsreichardt:
The issue was that `1.7.3 > 1.5.81` was `false` (should be `true`). Therefore, the app thought there was a newer version. I refactored `Version` (removing `_versionNameToInt`) to fix the issue.

Fixes #781

### DIFF
--- a/app/lib/pages/settings/changelog/change.dart
+++ b/app/lib/pages/settings/changelog/change.dart
@@ -52,19 +52,12 @@ class LineStringList with IterableMixin<String> {
 /// An Object which lets one compare different App-Versions.
 ///
 /// The Version must be three version number separated by docs (e.g. "1.3.1").
-class Version {
+class Version implements Comparable<Version> {
   final String name;
   static final _versionRegEx = RegExp(r'^(\d+\.)(\d+\.)(\d+)');
 
-  /// The major number of the version, incremented when making breaking changes.
   final int major;
-
-  /// The minor number of the version, incremented when adding new functionality
-  /// in a backwards-compatible manner.
   final int minor;
-
-  /// The patch number of the version, incremented when making
-  /// backwards-compatible bug fixes.
   final int patch;
 
   factory Version.parse({@required String name}) {
@@ -93,11 +86,11 @@ class Version {
   });
 
   bool operator >(Version other) {
-    return _compare(this, other) > 0;
+    return compareTo(other) > 0;
   }
 
   bool operator <(Version other) {
-    return _compare(this, other) < 0;
+    return compareTo(other) < 0;
   }
 
   bool operator >=(Version other) {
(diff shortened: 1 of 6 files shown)


app/lib/pages/settings/changelog/change.dart
```diff
@@ -50,21 +50,54 @@ class LineStringList with IterableMixin<String> {
...
+  final int major;
+
+  /// The minor number of the version, incremented when adding new functionality
+  /// in a backwards-compatible manner.
+  final int minor;
+
+  /// The patch number of the version, incremented when making
+  /// backwards-compatible bug fixes.
+  final int patch;
+
```
Jonas-Sander, line 69: Tbh we don't use the numbers like this. As it is just an app, not a library the numbers are just to show people how big of a change we made I guess. What even would be a breaking change? If the old versions don't work anymore? But I guess we would always make changes where old versions still work at least for some time.
I think that these comments might just cause more confusion as they state something that is not true in my opinion.

app/lib/pages/settings/changelog/change.dart
```diff
@@ -85,14 +118,21 @@ class Version {
...
+    if (a.major > b.major) return 1;
+    if (a.major < b.major) return -1;
+
+    if (a.minor > b.minor) return 1;
+    if (a.minor < b.minor) return -1;
+
+    if (a.patch > b.patch) return 1;
+    if (a.patch < b.patch) return -1;
+
+    return 0;
```
Jonas-Sander, line 136: Btw with `package:collection` you could write the code like this:

```suggestion
  int _compare(Version a, Version b) =>
      _compareMajor.then(_compareMinor).then(_comparePatch).call(a, b);

  int _compareMajor(Version a, Version b) => a.major.compareTo(b.major);
  int _compareMinor(Version a, Version b) => a.minor.compareTo(b.minor);
  int _comparePatch(Version a, Version b) => a.patch.compareTo(b.patch);
```

For me both ways are fine, just wanted to show you :)

app/test/settings/changelog/changelog_test.dart
```diff
@@ -39,16 +39,17 @@ void main() {
...
+    expect(Version.parse(name: "3.0.0") < Version.parse(name: "1.0.0"), false);
+    expect(Version.parse(name: "3.0.0") == Version.parse(name: "1.0.0"), false);
+    expect(Version.parse(name: "3.0.0") == Version.parse(name: "3.0.0"), true);
+    expect(Version.parse(name: "3.0.1") >= Version.parse(name: "3.0.0"), true);
+    expect(Version.parse(name: "3.0.0") >= Version.parse(name: "3.0.0"), true);
+    expect(Version.parse(name: "2.9.9") >= Version.parse(name: "3.0.0"), false);
+    expect(Version.parse(name: "2.9.9") <= Version.parse(name: "3.0.0"), true);
+    expect(Version.parse(name: "3.0.0") <= Version.parse(name: "3.0.0"), true);
+    expect(Version.parse(name: "3.0.1") <= Version.parse(name: "3.0.0"), false);
+    expect(Version.parse(name: "1.7.3") > Version.parse(name: "1.5.81"), true);
```
Jonas-Sander, line 52: Good job on adding the bug as a regression test case 👍

Jonas-Sander (approved):

app/lib/pages/settings/changelog/change.dart
nilsreichardt: Yes, you're right, I removed them 👍

app/lib/pages/settings/changelog/change.dart
nilsreichardt: Somehow this doesn't compile for me. With ChatGPT as a my second code review partner it found another smart way: Using the *Comparble` interface:

```dart
class Version implements Comparable<Version> {

  ...

  /// Compares this version to another [Version].
  ///
  /// Returns a negative value if this version is older than [other], a positive
  /// value if it's newer, and 0 if they're identical.
  @override
  int compareTo(Version other) {
    if (major != other.major) return major.compareTo(other.major);
    if (minor != other.minor) return minor.compareTo(other.minor);
    return patch.compareTo(other.patch);
  }
```

PR #786, opened 2023-08-22
Fix display of `showDisplaySettingsDialog` on privacy policy page

### LINKED ISSUE #785: Error when opening the "Anzeigeeinstellung" of the privacy policy page
image

nilsreichardt:
The issue was: The default value was `VisualDensitySetting.standard()` but this value was missing in the items list. Also adding a golden test to prevent this issue in the feature.

Fixes #785


nilsreichardt:
Golden test fails with 0.00% difference. I'm pretty sure this is the issue: https://github.com/flutter/flutter/issues/111739. Two possible solutions:

1. Add a threshold (requires a different package for the golden tests)
2. Using Codemagic for the golden tests (they offer M1 machines in the free plan)

I would choose the 2nd option to use Codemagic. This is the code snippet: https://github.com/flutter/flutter/issues/111739#issuecomment-1669215865

Jonas-Sander (approved):

PR #776, opened 2023-08-21
Replace `sign_in_with_apple` and `firebase_auth_oauth` with `firebase_auth`

nilsreichardt:
## Description

* Using the built-in social auth methods (removing `sign_in_with_apple` and `firebase_auth_oauth`)
* Replace the `firebaseAuthUser.reload()` strategy with `userChanges()` stream
  * Back then, `userChanges()` wasn't available what is the reason why we used `.reload()` after linking the Sharezone account with a sign in method
* Add support for Google Sign In on web in dev environment
* Users with Google Sign In doesn't pass their email address anymore (more GDPR friendly)

image

## Testing

Manually testing, if the following features are working:

* Sign In
* Link account with Apple/Google
* Delete account (need to reauthenticate)

| Provider | Android | iOS | macOS | Web |
|--------|--------|--------|--------|--------|
| Google Sign In | Works | Works | Works | Works |
| Apple Sign In | Works | Works | Works | Works |

## Related Tickets

Fixes #650
Fixes #649

PR #817, opened 2023-08-26
Fix linking account with email & password on iOS & macOS

### LINKED ISSUE #813: When linking the account with an email address and a password the local state doesn't update
https://github.com/SharezoneApp/sharezone-app/assets/24459435/78a0d7eb-8a25-4792-833e-5ef4c01cb435

The user is created in the backend. It's only a local state issue.

Reproducible with macOS but not with the latest alpah version on Android: 78972fe73cfb35164c57decca54a4308d31146c9

nilsreichardt:
## Description

The issue for #813 was that the `isAnonymous` attribute of the Firebase user has not been updated when linking an anonymous account with email address & password on iOS and macOS. The workaround is to use `reload()` on the Firebase auth user. The bug was caused by #776.

## Working Demo

https://github.com/SharezoneApp/sharezone-app/assets/24459435/c3646063-1179-4584-a789-025f12636c5b

## Related Tickets

Fixes #813

### DIFF
--- a/lib/authentification/authentification_base/lib/src/gateways/link_provider_gateway.dart
+++ b/lib/authentification/authentification_base/lib/src/gateways/link_provider_gateway.dart
@@ -46,13 +46,6 @@ class LinkProviderGateway extends BlocBase {
         EmailAuthProvider.credential(email: email, password: password);
     await userGateway.linkWithCredential(credential);
 
-    // Even when we are using the `userChanges()` stream, we still need to call
-    // `reload()` because the `isAnonymous` property is not updated on iOS (and
-    // macOS). When the bug is fixed, we can remove this call.
-    //
-    // Bug report: https://github.com/firebase/flutterfire/issues/11520
-    await userGateway.reloadUser();
-
     _analytics.logEmailAndPasswordLink();
 
     return true;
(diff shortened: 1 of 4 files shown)


lib/authentification/authentification_base/lib/src/gateways/link_provider_gateway.dart
```diff
@@ -46,6 +46,13 @@ class LinkProviderGateway extends BlocBase {
         EmailAuthProvider.credential(email: email, password: password);
     await userGateway.linkWithCredential(credential);
 
+    // Even when we are using the `userChanges()` stream, we still need to call
+    // `reload()` because the `isAnonymous` property is not updated on iOS (and
+    // macOS). When the bug is fixed, we can remove this call.
+    //
+    // Bug report: https://github.com/firebase/flutterfire/issues/11520
+    await userGateway.reloadUser();
+
```
Jonas-Sander, line 55: Couldn't the user be reloaded in the `linkWithCredential` method? In this way the code using the `userGateway` could stay "dumber".

Jonas-Sander (approved):

lib/authentification/authentification_base/lib/src/gateways/link_provider_gateway.dart
nilsreichardt: Yes, that's possible and alos shorter 👍

PR #833, opened 2023-08-26
Fix homework app bar for Google Pixel devices

nilsreichardt:
| Before | After |
|--------|--------|
| telegram-cloud-photo-size-2-5429174619237108870-y | telegram-cloud-photo-size-2-5429174619237108871-y |

### DIFF
--- a/app/lib/pages/homework/homework_dialog.dart
+++ b/app/lib/pages/homework/homework_dialog.dart
@@ -8,6 +8,7 @@
 
 import 'dart:async';
 import 'dart:developer';
+import 'dart:math' as math;
 
 import 'package:analytics/analytics.dart';
 import 'package:bloc_provider/bloc_provider.dart';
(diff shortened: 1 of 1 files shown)


app/lib/pages/homework/homework_dialog.dart
```diff
@@ -223,12 +223,27 @@ class _SaveButton extends StatelessWidget {
...
   Widget build(BuildContext context) {
-    return IconButton(
-      icon: const Text("SPEICHERN",
-          style: TextStyle(fontWeight: FontWeight.w600, color: Colors.white)),
-      tooltip: "Hausaufgabe speichern",
-      iconSize: 90,
-      onPressed: () => onPressed(context),
+    return Tooltip(
+      message: "Hausaufgabe speichern",
+      child: ElevatedButton(
```
nilsreichardt, line 228: If we want we to share the code of the buttons, not sure where to put the file for that and how to name it. Or we leave it like that.

Jonas-Sander: I mean it would make sense to share it because we mostly have the same format for the save/edit layouts for homework, blackboard and maybe events in the future.   
What about `/lib/sharezone_widgets`?

Jonas-Sander (approved):

Jonas-Sander (approved):

PR #846, opened 2023-08-28
Remove `--name-status` for Git diff check because otherwise you can't see what changed in the file

nilsreichardt:
Remove the `--name-status` option for the Git diff check because otherwise you can't see the changed content of the file, like in https://github.com/SharezoneApp/sharezone-app/actions/runs/6004428175/job/16286215041?pr=844

### DIFF
--- a/.github/workflows/cli_ci.yml
+++ b/.github/workflows/cli_ci.yml
@@ -114,12 +114,12 @@ jobs:
       # Therefore, fails this check if there are Git changes.
       - name: Fail if there are Git diffs
         run: |
+          # Fail if there are Git diffs and print the diff.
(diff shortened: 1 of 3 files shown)


Jonas-Sander:
Couldn't we run the command without the `--name-status` first and then the command with the flag? In this way you'd have a "summary" at the bottom. 
(Or first with the flag then without, idk what's better)

nilsreichardt:
> Couldn't we run the command without the `--name-status` first and then the command with the flag? In this way you'd have a "summary" at the bottom. (Or first with the flag then without, idk what's better)

Added ✅

Jonas-Sander (approved):

PR #876, opened 2023-09-04
Remove `package:optional`.

### LINKED ISSUE #749: Remove usage of `package:optional` from code.
After switching to null-safety, using `package:optional` doesn't make much sense any more.

Jonas-Sander:
We're now using nullable types instead of using `package:optional`.

Fixes #749


nilsreichardt (approved):
LGTM

PR #896, opened 2023-09-06
Use `flutter test` command for iOS integration test

nilsreichardt:


nilsreichardt:
Does not speed up the build...

PR #900, opened 2023-09-06
Add `golden_toolkit` to `sharezone_widgets`

nilsreichardt:


Jonas-Sander:
So this is only the scaffolding right? No additional tests will be run after this is merged, correct?

Jonas-Sander (approved):

nilsreichardt:
> So this is only the scaffolding right? No additional tests will be run after this is merged, correct?

Right, I needed it in #898 and a future PR

PR #906, opened 2023-09-07
Add `waitFor` method to make integration tests more stable

### LINKED ISSUE #905: Integration tests are failing when data are not immediately available
## Description

When the integration test is running on a device that is very slow or has a very slow internet connection, can the integration test fail because the test assumes that data are immediately available.

## Demo

https://github.com/SharezoneApp/sharezone-app/assets/24459435/5e1eccb5-9030-40d7-91f5-21e9c5f8dfff

## Steps to reproduce

1. Set the Android Emulator internet to a very slow connection
2. Log in
3. Open group page

image

nilsreichardt:
#### Summary:
In this pull request, we have introduced a new utility function `waitFor` to enhance the stability of our integration tests by ensuring certain widgets are present in the widget tree before proceeding with subsequent test operations. This alleviates the issue of tests failing due to widgets not being loaded within the expected time frame.

## Demo

https://github.com/SharezoneApp/sharezone-app/assets/24459435/2039e579-18b4-4f53-af3c-d868d6abb934

## Related tickets

Fixes #905
Workaround for https://github.com/flutter/flutter/issues/88765

### DIFF
--- a/app/integration_test/app_test.dart
+++ b/app/integration_test/app_test.dart
@@ -161,11 +161,6 @@ class _UserCredentials {
 /// ```dart
 /// await waitFor(tester, find.text('Hello, World'));
 /// ```
-///
-/// ## Exceptions
-///
-/// Throws an [Exception] if the [timeout] is reached before the widget is
-/// found.
 Future<void> waitFor(
   WidgetTester tester,
   Finder finder, {
--- a/app/integration_test/app_test.dart
+++ b/app/integration_test/app_test.dart
@@ -85,14 +85,17 @@ void main() {
       await tester.tap(find.byKey(const Key('nav-item-group-E2E')));
       await tester.pumpAndSettle();
 
+      // Ensure that the group list is loaded.
+      await waitFor(tester, find.text('Meine Klasse:'));
+
       // We assume that the user is in at least 5 groups with the following
       // group names.
-      await waitFor(tester, find.text('10A'));
-      await waitFor(tester, find.text('Deutsch LK'));
-      await waitFor(tester, find.text('Englisch LK'));
-      await waitFor(tester, find.text('Französisch LK'));
-      await waitFor(tester, find.text('Latein LK'));
-      await waitFor(tester, find.text('Spanisch LK'));
+      expect(find.text('10A'), findsOneWidget);
+      expect(find.text('Deutsch LK'), findsOneWidget);
+      expect(find.text('Englisch LK'), findsOneWidget);
+      expect(find.text('Französisch LK'), findsOneWidget);
+      expect(find.text('Latein LK'), findsOneWidget);
+      expect(find.text('Spanisch LK'), findsOneWidget);
 
       print("Test: User should be able to load timetable");
       await tester.tap(find.byKey(const Key('nav-item-timetable-E2E')));
--- a/app/integration_test/app_test.dart
+++ b/app/integration_test/app_test.dart
@@ -87,12 +87,12 @@ void main() {
 
       // We assume that the user is in at least 5 groups with the following
       // group names.
-      expect(find.text('10A'), findsOneWidget);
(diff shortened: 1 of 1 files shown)


app/integration_test/app_test.dart
```diff
@@ -134,3 +134,46 @@ class _UserCredentials {
...
+  Duration timeout = const Duration(seconds: 70),
+}) async {
+  final end = tester.binding.clock.now().add(timeout);
+
+  do {
+    if (tester.binding.clock.now().isAfter(end)) {
+      throw Exception('Timed out waiting for $finder');
+    }
+
+    await tester.pumpAndSettle();
```
Jonas-Sander, line 176: Wouldn't this hang indefinitely if there is an endless/repeating Animation playing?

Jonas-Sander:
I assume you made the PR description with AI? I don't like it, it's way too much text and 2. and 4. are pretty much the same

app/integration_test/app_test.dart
```diff
@@ -87,12 +87,12 @@ void main() {
 
       // We assume that the user is in at least 5 groups with the following
       // group names.
-      expect(find.text('10A'), findsOneWidget);
-      expect(find.text('Deutsch LK'), findsOneWidget);
-      expect(find.text('Englisch LK'), findsOneWidget);
-      expect(find.text('Französisch LK'), findsOneWidget);
-      expect(find.text('Latein LK'), findsOneWidget);
-      expect(find.text('Spanisch LK'), findsOneWidget);
+      await waitFor(tester, find.text('10A'));
```
Jonas-Sander, line 90: So one can't specify sth like `findsOneWidget` with the new method?

Jonas-Sander: Also couldn't we use the `waitFor` for only the first group name since the others are loaded by then? Otherwise it might take until we hit the timeout maybe several times if there are a few groups where the account is missing.


In this way there would also be clear separation that we `waitFor` the groups being loaded and then use `findsOneWidget` so that it is clear that we expect one group with the name for all the groups.

app/integration_test/app_test.dart
Jonas-Sander: Since it's `do while` wouldn't it mean that with an endless animation we would never even come to the `finder.evaluate()` part since it would hang at `finder.pumpAndSettle`?

nilsreichardt: Yes, in the case we have an endless animation, we can't use `tester.pumpAndSettle()` for this place in the app anymore

app/integration_test/app_test.dart
nilsreichardt: We can first check if can find `Meine Klasse:`:

```dart
// Ensure that the group list is loaded. When the school class is loaded,
// we assume that the courses list is loaded as well.
await waitFor(tester, find.text('Meine Klasse:'));
```

and then test with `expect` again as we did before.

app/integration_test/app_test.dart
nilsreichardt: > Since it's do while wouldn't it mean that with an endless animation we would never even come to the finder.evaluate()

Even without a `while` loop and just `tester.pumpAndSettle()` would an endless animation never complete the `tester.pumpAndSettle` (only when the default timeout of 10 minutes for `pumpAndSettle`() hits).

Jonas-Sander: I mean since it works for this we can merge it like this if u want, but couldn't we just replace `pumpAndSettle`  with `pump` since its looped anyways?

nilsreichardt: Yes, this could work. Locally it worked. Let's try this.

nilsreichardt:
> Btw, what do you think of this?

Yes, looks great. I updated the code 👍

PR #910, opened 2023-09-07
Add alias `lh` for `sz license-headers` command

nilsreichardt:
Often, I'm too stupid to write "license-headers" and I need a couple of seconds to type it correctly. Having an alias like `lh` would be make it much easier to write `sz lh add` to add the license headers.

### DIFF
--- a/tools/sz_repo_cli/lib/src/commands/src/license_headers_command.dart
+++ b/tools/sz_repo_cli/lib/src/commands/src/license_headers_command.dart
@@ -14,4 +14,7 @@ class LicenseHeadersCommand extends Command {
 
   @override
   String get name => 'license-headers';
+
+  @override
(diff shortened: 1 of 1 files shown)


Jonas-Sander (approved):

tools/sz_repo_cli/lib/src/commands/src/license_headers_command.dart
```diff
@@ -14,4 +14,7 @@ class LicenseHeadersCommand extends Command {
 
   @override
   String get name => 'license-headers';
+
+  @override
+  List<String> get aliases => ['lh'];
```
Jonas-Sander, line 19: Very cool, didn't know about this! Thanks for sharing.

PR #911, opened 2023-09-07
Add signing steps to the `sz deploy ios` command

nilsreichardt:
Currently, we run the signing steps in the GitHub Actions file. However, we can't do this for the macOS release because signing is part of the deploy process. As prework, I added the signing steps to the iOS command because the signing for iOS is easier than for macOS. When this works, can we make a new PR for the `sz deploy macos` command.

Manually tested it with: https://github.com/SharezoneApp/sharezone-app/actions/runs/6119211836/job/16608705828


tools/sz_repo_cli/lib/src/commands/src/deploy_ios_command.dart
```diff
@@ -94,38 +95,51 @@ class DeployIosCommand extends Command {
...
+      await publishToAppStoreConnect(
+        appStoreConnectConfig: appStoreConnectConfig,
+        stage: argResults![releaseStageOptionName] as String,
+        whatsNew: argResults![whatsNewOptionName] as String?,
+        path: 'build/ios/ipa/*.ipa',
+        repo: _repo,
+        stageToTracks: _iosStageToTracks,
+      );
+    } finally {
+      await keychainUseLogin();
```
Jonas-Sander, line 141:
```suggestion
      // Fixes potential authentication issues after running keychain commands.
      // Only really necessary when running on local machines.
      await keychainUseLogin();
```

I  think we can add this here as well to save ourselves the "hop" to the documentation for this comment in the future.

Jonas-Sander (approved):
Nice, LGTM

PR #928, opened 2023-09-08
Remove code for meetings feature

nilsreichardt:
Meetings are not available anymore, and the message about the shutdown was displayed long enough.


app/lib/groups/analytics/group_analytics.dart
```diff
@@ -10,18 +10,11 @@ import 'package:analytics/analytics.dart';
 import 'package:bloc_base/bloc_base.dart';
 
 class GroupAnalytics extends BlocBase {
+  // ignore: unused_field
```
nilsreichardt, line 13: We need to leave the `GroupAnalytics` class in the code base because of #930

Jonas-Sander (approved):

PR #929, opened 2023-09-08
Get plus fallback price from remote config.

### LINKED ISSUE #908: Load fallback price for Sharezone Plus from Firebase Remote Config

Jonas-Sander:
Displays a fallback price from Remote Config if we can't load price from RevenueCat.  
Displays a hardcoded fallback price if we can't load the fallback price from Remote Config.

Fixes #908


Jonas-Sander:
@nilsreichardt Some open questions I faced when implementing this:

Might the prices for students, teachers and parents differ in the future? Then we probably need three distinct fallback prices, don't we?

Also wouldn't there be the chance of a mismatch between the actual price and the fallback price? Since we try to load the RevenueCat price only once, there might be the chance that it fails, we display the fallback and the user tries buying it (RevenueCat now functions again) and then he will be asked for another price. This could happen with RemoteConfig (we forgot to update the price or the new config isn't on the users device yet) or with the hardcoded fallback price (might be outdated). I'm a bit cautious because this might be legal trouble for us to first display price A and then actually ask for price B.   

So what would be the best way to go forward with this? Create a Stream for the Sharezone Plus price (try loading from RevenueCat repeatedly) and include if this is a fallback price? If it is a fallback price then deactivate the buy button?  
Would this be worth the effort? We could also just display an error/show loading indicator if the price can't be fetched and don't have any kind of fallback price.

nilsreichardt:
My idea with the fallback price was to primarily use this on platforms where RevenueCat isn't available (macOS and Web). Still showing the price and a buy button would allow to track how many people would buy on these platforms.

For Android & iOS, we can use the retry method of the retry package and if this doesn't work just show an error message. If the value isn't loaded, we can show a shimmer loading animation.

For parents, students and teachers we can use different remote config values, if this isn't too much work.

PR #930, opened 2023-09-08
Add more group analytics

nilsreichardt:


app/lib/groups/analytics/group_analytics.dart
```diff
@@ -8,28 +8,98 @@
...
 
   @override
   void dispose() {}
 }
 
 class GroupEvent extends AnalyticsEvent {
-  GroupEvent(String name) : super('group_$name');
-}
-
-class GroupSettingsEvent extends GroupEvent {
```
nilsreichardt, line 33: I remove the `GroupSettingsEvent` class because with a `_settings' prefix in the analytics name we easily hit the 32 character limit for an analytics name.

Jonas-Sander (approved):

PR #940, opened 2023-09-12
Add support for line breaks when starting a new version from GitHub Actions

nilsreichardt:
The current problem was that when passing `\n` for a new line, `\n` was escaped.


nilsreichardt:
Moved to the action files.

Jonas-Sander (approved):
Thanks, I like this way better 👍

PR #945, opened 2023-09-13
Restrict access for filter by school class for Sharezone Plus users

### LINKED ISSUE #852: Restrict the selection of school classes in the timetable to Sharezone Plus users only
## Design

https://www.figma.com/file/EIlLPe7KdF4bQXLQVAsDeM/SharezoneApp?type=design&node-id=2372-4454&mode=design&t=sRCMFs02Gv9dau1D-4

Behavior:

* When tapping as free user on a school class, redirect to the Sharezone Plus page

nilsreichardt:
## Description

This PR restricts access to the school class filters for Sharezone Plus users. At the moment it just redirects to the Sharezone Plus page if the user doesn't have Sharezone Plus. But I want to show a dialog with a video demo first, so that the user can verify "Ah, yes, that's the feature for what I'm looking for". Will be implemented in a future PR (#953).

## Demo

Desktop:

https://github.com/SharezoneApp/sharezone-app/assets/24459435/42c76e9f-79eb-482f-a1f7-65f8ec09a73a

Mobile:

image

## Related Tickets

Closes #852


app/lib/timetable/timetable_page/school_class_filter/school_class_filter.dart
```diff
@@ -96,11 +100,29 @@ class _SchoolClassFilterBottomBarState
...
       bloc.changeSchoolClassFilter(selectedOption);
     }
   }
 
+  bool _shouldOpenSharezonePlusPage(
+      BuildContext context, SchoolClassFilter selectedOption) {
+    final isPlusEnabled = context.read<SubscriptionEnabledFlag>().isEnabled;
+    final hasPlus = context.read<SubscriptionService>().isSubscriptionActive();
+    return selectedOption.isPlusRequiredFilter && !hasPlus && isPlusEnabled;
+  }
```
Jonas-Sander, line 119:
```suggestion
  bool _shouldOpenSharezonePlusPage(
      BuildContext context, SchoolClassFilter selectedOption) {
    final isUnlocked = context
        .read<SubscriptionService>()
        .hasFeatureUnlocked(SharezonePlusFeature.filterTimetableByClass);
    return selectedOption.shouldFilterForClass && !isUnlocked;
  }
```

We already implemented `SubscriptionService` in such a way, that if Sharezone Plus is not activated then all features are unlocked.

I think for the change above you would need to create `SharezonePlusFeature.filterTimetableByClass` and maybe even a new `SubscriptionTier`?

Also I removed the `SchoolClassFilter.isPlusRequiredFilter` which I mean is not wrong, but I felt a bit uneasy with this (I'm not sure if the logic should be on the `SchoolClassFilter`). This is just a hunch though, you are free to keep it there.

app/lib/timetable/timetable_page/school_class_filter/school_class_filter.dart
```diff
@@ -119,28 +141,41 @@ class _SchoolClassFilterBottomBarState
     );
   }
 
+  PopupMenuItem<SchoolClassFilter> schoolClassToItem(
```
Jonas-Sander, line 144: Nit: You don't have to change it
```suggestion
  PopupMenuItem<SchoolClassFilter> schoolClassToFilterItem(
```

Jonas-Sander (changes requested):

app/lib/timetable/timetable_page/school_class_filter/school_class_filter.dart
nilsreichardt: Ah, I forgot about `hasFeatureUnlocked` 😂

Jonas-Sander (approved):
Nice, LGTM 👍

PR #956, opened 2023-09-14
Remove adding the stage to build name on iOS and macOS because Flutter doesn't support this.

nilsreichardt:
Unfortunately, we can't add `alpha` and `beta` as suffix to the build name because Flutter doesn't support it on iOS and macOS.

See:
* https://github.com/flutter/flutter/issues/27589#issuecomment-573121390
* https://github.com/flutter/flutter/issues/115483


Jonas-Sander (approved):
I think I saw some projects who have a `const String currentVersion = "1.4.8"` or something similar in the Dart code. Wouldn't this be a possibility for the future?

Maybe sth like this:
```dart
//autogenerated
const String currentBaseVersion = `1.4.8`;
String get currentVersion {
if developmentStage == beta {
 return "$currentBaseVersion-beta";
 }
return currentBaseVersion;
}
```

nilsreichardt:
You mean when we show the version in the app, like on the about us page and adding manually the stage? Yes, this would be possible 👍

PR #962, opened 2023-09-14
Add Sharezone Plus support

### LINKED ISSUE #885: Add plus support to support page
Design: https://www.figma.com/file/EIlLPe7KdF4bQXLQVAsDeM/SharezoneApp?type=design&node-id=2380-5246&mode=design&t=XAyp3Z0WOjT8vZ58-4

nilsreichardt:
## Description

This adds the plus support to our support page.

### Challenges

When implementing this feature, I had the following challenge: We also use the support page in areas of the app, where the user isn't signed in and can't buy Sharezone Plus and `SubscriptionService` and `SubscriptionEnabledFlag` aren't available.

I solved it by:

* Initializing the `SupportPageController` at two different locations: `SharezoneApp` (user is authenticated) and `AuthApp` (user is not authenticated)
  * Therefore, I can set the attributes of the `SupportPageController` to `Stream.value(false)`
* I moved `SubscriptionEnabledFlag` above `SharezoneApp` and `AuthApp` to be able to access it in both subtrees.  

## Demo

image

## Related Tickets

Closes #885


nilsreichardt:
@Jonas-Sander Can you already review the PR? If this approach is okay, I can also add a few widgets tests for the logic.

Jonas-Sander:
grafik

I think these could have a bit more padding

Jonas-Sander:
grafik

It looks like the gap is bigger between the 1st and 2nd than between the 2nd and 3rd

Jonas-Sander (approved):

nilsreichardt:
> I think these could have a bit more padding

Yes, I thought the same but after 10 minutes trying to fix it, I gave up because the problem is that the padding should only be applied when it breaks into two lines and not always. If you have an idea how to fix it, let me know. If not, let's keep like this, we can fix it later.

Jonas-Sander (approved):
LGTM

PR #965, opened 2023-09-14
Remove `SharezoneContext` from `MatchingTypeOfUserStreamBuilder`

nilsreichardt:
Should rename this widget since we are not using a `StreamBuilder` anymore?


Jonas-Sander (approved):
Yes, please rename (also the file). Maybe just keep it like before but remove "Stream"? (`MatchingTypeOfUserBuilder`)  
I don't have strong opinion for the new name.

PR #979, opened 2023-09-15
Use `package:bloc` for Sharezone Plus page.

Jonas-Sander:
https://github.com/SharezoneApp/sharezone-app/assets/29028262/a52ca677-5159-4376-a879-4d65e9e571ca

### DIFF
--- a/app/lib/blocs/sharezone_bloc_providers.dart
+++ b/app/lib/blocs/sharezone_bloc_providers.dart
@@ -23,6 +23,7 @@ import 'package:firebase_hausaufgabenheft_logik/firebase_hausaufgabenheft_logik_
 import 'package:firebase_messaging/firebase_messaging.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_bloc/flutter_bloc.dart' as bloc_lib;
 import 'package:hausaufgabenheft_logik/hausaufgabenheft_logik_lehrer.dart';
 import 'package:hausaufgabenheft_logik/hausaufgabenheft_logik_setup.dart';
 import 'package:holidays/holidays.dart' hide State;
@@ -315,6 +316,8 @@ class _SharezoneBlocProvidersState extends State<SharezoneBlocProviders> {
       isSubscriptionEnabledFlag: subscriptionEnabledFlag,
     );
 
+    final revenueCatPurchaseService = RevenueCatPurchaseService();
+
     // In the past we used BlocProvider for everything (even non-bloc classes).
     // This forced us to use BlocProvider wrapper classes for non-bloc entities,
     // Provider allows us to skip using these wrapper classes.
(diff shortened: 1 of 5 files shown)


Jonas-Sander:
The code is still kinda shitty, but I wanted to know what you think of using `package:bloc` for this? (I know I said lets use `Provider`).

@nilsreichardt

nilsreichardt:
I think the biggest thing is here that you are using sealed classes but the structure could also easily be achieved with `provider` or `riverpod` (see #983) or what would be the benefit of `bloc` here?

Jonas-Sander:
> I think the biggest thing is here that you are using sealed classes but the structure could also easily be achieved with `provider` or `riverpod` (see #983) or what would be the benefit of `bloc` here?

Yes thats true, there is no hard benefit. One theoretical benefit could be that there is an `BlocObserver` class available which could aid in debugging or logging (altough I don't really know if it would help).     

I like the `state` approach with the `SharezonePlusPageController2`, so maybe lets go that way?   

I'm not so sure about the `riverpod` approach, although this might be my unfamiliarity - I dislike that one has to change the widget type to be a `ConsumerWidget`. Also it feels like too much magic to me, but again this might just be my unfamiliarity.

nilsreichardt:
> Yes thats true, there is no hard benefit. One theoretical benefit could be that there is an BlocObserver class available which could aid in debugging or logging (altough I don't really know if it would help).

Yes, that's something I thought as well and it's actually an argument because it adds a lot of valuable logging without actively adding logging. Haven't found a good way to do this with `provider`. `riverpod` also has an [observer feature](https://riverpod.dev/docs/concepts/provider_observer) like `bloc`.

> I like the state approach with the SharezonePlusPageController2, so maybe lets go that way?

Yes, I think so

> I'm not so sure about the riverpod approach, although this might be my unfamiliarity - I dislike that one has to change the widget type to be a ConsumerWidget. Also it feels like too much magic to me, but again this might just be my unfamiliarity.

I really like `riverpod` and have the feeling that it's the most used state management solution for new projects (no data to prove it, just personal feeling from talking to other engineers) but now using it for Sharezone is a bit difficult because it only works good when you only use `riverpod`. `bloc` and `provider` can easily be used without being forced to use it everywhere. Therefore, I would either use `bloc` or `provider`.

Jonas-Sander:
> because it only works good when you only use riverpod

Yes that was one thing I thought as well when you commented in the other PR about the `BuildContext`, but I wasn't sure enough to comment.  

I think there might(!) be good arguments to maybe use one simple state management solution for feature A and a more complex one for feature B. Not saying that we necessarily have to do that but in this case using a state management that "spreads" is a bit counter productive.

I think I should really play with `riverpod` though, since it's used so much. Maybe I'll like it.

nilsreichardt:
> I think there might(!) be good arguments to maybe use one simple state management solution for feature A and a more complex one for feature B.

Yes, I'm open to use several state managements (like `provider` and `bloc`) as long it makes the code more simple when you use the more suitable one for the feature

PR #986, opened 2023-09-16
Add advantage "see homework done list" to Sharezone Plus page

nilsreichardt:
image


app/test_goldens/sharezone_plus/sharezone_plus_page_test.dart
```diff
@@ -91,6 +92,7 @@ void main() {
             ChangeNotifierProvider<SharezonePlusPageController>(
               create: (context) => controller,
             ),
+            Provider<TypeOfUser?>.value(value: TypeOfUser.student),
```
nilsreichardt, line 95: For now, I would hard-code it `TypeOfUser.student`. Later, we should create different goldens for each `TypeOfUser`.

Jonas-Sander (approved):

PR #990, opened 2023-09-17
Add basic widget tests for Sharezone plus page (plus small other changes).

Jonas-Sander:
I added some basic widget/UI tests by using a mock `SharezonePlusPageController`.

Additional changes: 
* If the user is subscribed we show the current price.
* Add fake loading, subscribing and canceling.

The overall thing is still in progress, I just wanted to commit a working interim result.


Jonas-Sander:
I decided to not use `package:bloc` for now. If I think it's warranted when developing the page further then I might add it.

app/lib/sharezone_plus/page/sharezone_plus_page_controller.dart
```diff
@@ -33,50 +35,38 @@ class SharezonePlusPageController extends ChangeNotifier {
...
 
-    _hasPlusSubscription = hasPlus.listen((hasPlus) {
-      this.hasPlus = hasPlus;
-      notifyListeners();
-    });
-  }
+  /// The price for the Sharezone Plus per month, including the currency sign.
+  ///
+  /// If the user is subscribed to Sharezone Plus then this is the price for
+  /// his current subscription.
```
nilsreichardt, line 52: FYI: It's likely that we are not going to support this for the beginning because we don't know which subscription the user has when the user is on a different platform, e.g. on the web.

nilsreichardt: Ok, it's easier than I thought to implement because RevenueCat includes this information into the Firestore document. But still something we haven't implemented on the backend site and it's not a blocker to launch from my side so probably we ship later as an update.

nilsreichardt (approved):
LGTM

app/lib/sharezone_plus/page/sharezone_plus_page_controller.dart
Jonas-Sander: Let's keep it for now, if it takes more than an hour to implement we can still remove it later.

PR #1013, opened 2023-09-18
Enable `cache` when installing Flutter in CI

nilsreichardt:


nilsreichardt:
Seems so that cache is automatically enabled.

PR #1019, opened 2023-09-19
Add `--rollout-percentage` to `sz deploy android`

nilsreichardt:
This PR adds the `--rollout-percentage` to `sz deploy android`: The percentage (between 0.0 - 1.0) of the user fraction when uploading to the rollout track (setting to 1 will complete the rollout).

Additionally, this PR adds the parameters to `stable.yml` (default: 0.2) and `beta.yml` (default: 1.0).

### DIFF
--- a/.github/workflows/beta.yml
+++ b/.github/workflows/beta.yml
@@ -36,6 +36,12 @@ on:
           Android changelog: Used for the Google Play Store release notes. If the 
           changelog is not provided, the deployment for Android will be skipped.
           Use '\n' for line breaks.
+      android-rollout-percentage:
+        type: number
+        description: |
+          Android rollout percentage: The percentage of users that should receive
+          the new version.
+        default: '1.0'
 
 # Set permissions to none.
 #
@@ -205,4 +211,5 @@ jobs:
 
           sz deploy android \
             --stage beta \
-            --whats-new "$CHANGELOG_WITH_NEW_LINES"
+            --whats-new "$CHANGELOG_WITH_NEW_LINES" \
+            --rollout-percentage ${{ github.event.inputs.android-rollout-percentage }}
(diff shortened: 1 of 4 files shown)


Jonas-Sander:
To be honest I find that behavior might be a bit surprising.  
If we want to do that I think at least it:
* should be printed in the console that a 20% rollout is done
* the url to the PlayStore Console site where one can adjust/increase the rollout should be printed in the console
(If we trigger it from GitHub manually

> For now, I hard-coded the 20% for stable. Later on, we might add an "Is hot fix?" bool for our stable.yml and beta.yml which roll-outs to 100% immediately. As long, we don't have this, we can manually increase the rollout from 20% to 100% in the Google Play Console.

Why not use a `--rollout-percentage` flag? Through that one could even supply the value from GitHub Actions. I'd leave it to your choice if should still default to 20% rollout for stable in this case. I think this is the better alternative tbh.

Also just food for thought for the future: We could make a `rollout` a seperate concept in the CLI in the future. `sz rollout android change --rollout-percentage 40`. iOS doesn't seem support the same granularity, but still a phased approach which could be represented in the CLI (e.g. show rollout percentage, pause rollout).

nilsreichardt:
> Why not use a --rollout-percentage flag? Through that one could even supply the value from GitHub Actions. I'd leave it to your choice if should still default to 20% rollout for stable in this case. I think this is the better alternative tbh.

Yes, thanks for the suggestion. This is the best option 👍 

> Also just food for thought for the future: We could make a rollout a seperate concept in the CLI in the future. sz rollout android change --rollout-percentage 40. iOS doesn't seem support the same granularity, but still a phased approach which could be represented in the CLI (e.g. show rollout percentage, pause rollout).

Yes, this would be a great feature for the future because I really like the phased rollout that offers App Store Connect. However, I think it will take a long moment until I will build this because for now, I will use increase it manually via the Google Play Console.

Jonas-Sander (approved):
I like this way more 👍 LGTM

Would it be possible to add this? 

> * should be printed in the console that a x% rollout is being done
> * the url to the PlayStore Console site where one can adjust/increase the rollout should be printed in the console

PR #1042, opened 2023-09-20
Add option to view past events for Sharezone Plus users

### LINKED ISSUE #242: A user should be able to view the events from the past.
## Description
Currently, it's only possible to view the calendar events from today or of the future. But sometimes it's helpful to view the events from the past. Possible use cases are:
* You need to take a make-up exam (Klausur nachschreiben) and you want to view the topics of the exam
* You want to view the topics of the last exams just for fun

## Implementation
The implementation should be straightforward because we just need to change this query to `isLessThanOrEqualTo`:
https://github.com/SharezoneApp/sharezone-app/blob/4c28345ab6c67c7be9f9a615cc5b217bde911f7c/app/lib/util/api/timetableGateway.dart#L99

### LINKED ISSUE #891: Add option to view past events
Design: https://www.figma.com/file/EIlLPe7KdF4bQXLQVAsDeM/SharezoneApp?type=design&node-id=2380-5868&mode=design&t=VU5sZUg4J4TkzptR-4

nilsreichardt:
## Description

This PR adds the feature to view the past events.

I haven't added lazy loading to avoid complexity. If this feature is actually used, we can add lazy later.

## Demo

Unlocked:

https://github.com/SharezoneApp/sharezone-app/assets/24459435/8db4c9e4-e26b-4150-be42-96cfd32bf8e9

Not unlocked:

image

## Related Tickets

Closes #242
Closes #891


app/lib/calendrical_events/provider/past_calendrical_events_page_controller.dart
```diff
@@ -0,0 +1,162 @@
...
+        _listenToPastEvents();
+      } else {
+        state = PastCalendricalEventsPageNotUnlockedState();
+        notifyListeners();
+      }
+    })
+      ..onError((e) {
+        _setError('$e');
+      });
+  }
```
Jonas-Sander, line 65: Since the `PastCalendricalEventsPageController` is always newly created when the page is viewed, couldn't we just access `.hasFeatureUnlocked` instead of `.hasFeatureUnlockedStream`?

app/lib/calendrical_events/provider/past_calendrical_events_page_controller.dart
```diff
@@ -0,0 +1,162 @@
...
+import 'package:sharezone/sharezone_plus/subscription_service/subscription_service.dart';
+import 'package:sharezone/timetable/src/widgets/events/event_view.dart';
+import 'package:sharezone/util/api/course_gateway.dart';
+import 'package:sharezone/util/api/school_class_gateway.dart';
+import 'package:sharezone/util/api/timetable_gateway.dart';
+
+class PastCalendricalEventsPageController extends ChangeNotifier {
+  late PastCalendricalEventsPageState state;
+  EventsSortingOrder _sortingOrder = EventsSortingOrder.descending;
+  EventsSortingOrder get sortingOrder => _sortingOrder;
```
Jonas-Sander, line 26: Feels a bit weird to have `sortingOrder` state here but also a `state` property.  
But also I have the feeling that doing it otherwise (include `PastCalendricalEventsPageState .sortingOrder`) might make it more complex to handle changing the sorting while e.g. loading/having an error. I'm not sure.  

So either:
1. Rename `state`, my idea would be `.eventsState` 
2. Add the `sortingOrder` to each of the `PastCalendricalEventsPageState` and remove from the controller.

Jonas-Sander (approved):

app/lib/calendrical_events/provider/past_calendrical_events_page_controller.dart
nilsreichardt: Yes, true 👍

PR #1068, opened 2023-09-21
Upgrade Sharezone to v1.7.8

nilsreichardt:


nilsreichardt:
Merging this without a review since otherwise all the macOS deployments fail:

https://github.com/SharezoneApp/sharezone-app/actions/runs/6260397107/job/16998351952

PR #1070, opened 2023-09-21
Add temporary remote config values for testing.

Jonas-Sander:
I want to test how remote config behaves if the website is hosted on a static domain (`alpha.web.sharezone.net`):
* Behavior on first load
* Behavior after completely closing the browser and opening the same site again
* Behavior when refreshing the page

I just used some of the values that are already in prod remote config but are not used anymore.

This is important so we know how a fallback price for Sharezone Plus would behave.

### DIFF
--- a/app/lib/sharezone_plus/page/sharezone_plus_page.dart
+++ b/app/lib/sharezone_plus/page/sharezone_plus_page.dart
@@ -38,9 +38,15 @@ class SharezonePlusPage extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final remoteConfig = getRemoteConfiguration();
-    log('revenuecat_api_key: ${remoteConfig.getString('revenuecat_api_key')}');
-    log('zeige_pilotschule_karte: ${remoteConfig.getString('zeige_pilotschule_karte')}');
-    log('useCfHolidayEndpoint: ${remoteConfig.getBool('useCfHolidayEndpoint')}');
+    // ignore: avoid_print
+    print(
+        'revenuecat_api_key: ${remoteConfig.getString('revenuecat_api_key')}');
+    // ignore: avoid_print
+    print(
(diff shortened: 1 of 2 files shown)


app/lib/sharezone_plus/page/sharezone_plus_page.dart
```diff
@@ -34,6 +35,14 @@ class SharezonePlusPage extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final remoteConfig = getRemoteConfiguration();
+    print(
+        'revenuecat_api_key: ${remoteConfig.getString('revenuecat_api_key')}');
+    print(
+        'zeige_pilotschule_karte: ${remoteConfig.getString('zeige_pilotschule_karte')}');
+    print(
+        'useCfHolidayEndpoint: ${remoteConfig.getBool('useCfHolidayEndpoint')}');
```
Jonas-Sander, line 44: This one is not defined in the defaults and only on the remote config server (locally it should still print false because of the default fallback values of remote config, but lets see)

nilsreichardt (approved):
LGTM

(You could also use the web preview URL of this PR and then the sharezone-debug project without merging)

Jonas-Sander:
| Action | Has remote/updated values|
|---|---|
| First open (Incognito Tab) | ❌ |
| Closing tab and re-opening (Not leaving Incognito) | ✅ |
| Reloading page | ✅ |
| Closing Incognito completely and re-opening | ❌ |
| Logging out and logging back in with same account | ❌ |
| Logging out and logging back in with other account | ❌ |
| Closing Tab and reopen (Not leaving incognito) and login with another account | ✅ |

Jonas-Sander:
@nilsreichardt So I guess we would have to wait for the values to be loaded on web and not use the "Load new values for next startup"-strategy if we plan to use fallback values from remote config. WDYT?

PR #1087, opened 2023-09-24
Pre-fill user ID, name and email when making an appointment for the Sharezone Plus video call support

nilsreichardt:
## Demo

https://github.com/SharezoneApp/sharezone-app/assets/24459435/0a2e305a-ec54-46fb-bac2-7e4b2b87a9bd

### DIFF
--- a/app/lib/support/support_page_controller.dart
+++ b/app/lib/support/support_page_controller.dart
@@ -69,7 +69,8 @@ class SupportPageController extends ChangeNotifier {
   /// ```
   /// https://sharezone.net/sharezone-plus-video-call-support?userId=userId123&name=My Cool Name&email=my@email.com
   /// ```
-  /// Note: The URL returned by this method should be encoded before usage.
+  /// Note: The URL is not encoded because the `launchURL` method will encode
+  /// the URL.
   ///
   /// This method throws [UserNotAuthenticatedException] if the `userId` is
   /// `null`.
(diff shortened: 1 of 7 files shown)


app/lib/support/support_page_controller.dart
```diff
@@ -42,11 +61,66 @@ class SupportPageController extends ChangeNotifier {
...
 
+  /// Returns the unencoded URL for the video call appointments page, with
+  /// prefilled user ID, user name, and user email (for non-private Apple
+  /// email).
+  ///
+  /// Example URL:
+  /// ```
+  /// https://sharezone.net/sharezone-plus-video-call-support?userId=userId123&name=My Cool Name&email=my@email.com
+  /// ```
+  /// Note: The URL returned by this method should be encoded before usage.
```
Jonas-Sander, line 72: Why is the URL unencoded? You don't explicitly encode it afterwards, is it done by `launchURL(url, context: context);`? This should be clarified in the documentation.

Jonas-Sander (approved):

app/lib/support/support_page_controller.dart
nilsreichardt: Yes, it's done by the `launchURL` method

nilsreichardt: I added "Note: The URL is not encoded because the `launchURL` method will encode the URL."

PR #1104, opened 2023-10-01
Replace `FocusManager` call with `focusNode.requestFocus` in `delayKeyboard(...)`.

Jonas-Sander:
The old way didn't work for me to request focus for the title field in the add homework dialog. This fixes it.


nilsreichardt (approved):
LGTM

Btw: This will also fix the issue for the blackboard dialog 👍

PR #1112, opened 2023-10-04
Refactor some of the homework dialog code.

Jonas-Sander:
This PR splits up the "is private" and "are submissions enabled" parts of the homework dialog into bloc-agnostic (`xyzBase` widgets) and composing widgets.
Doing this should make changing the homework dialog easier in the future. Also if we ever decide to break the edit and create dialog up into different widgets we need to use the bloc-agnostic base widgets anyways.

Additionally I moved the `NextLessonCalculator` out of the `HomeworkDialogApi` and made its attributes private. This should ease mocking in future widget tests of the homework dialog.


nilsreichardt (approved):
LGTM

Issue #1115, opened 2023-10-06
Markdown analytics are wrong when editing a homework.


Jonas-Sander:
## Describe the bug
When editing a homework and not using markdown in the new description a "`markdown_used`" analytics event is logged regardless.

## Steps to reproduce the bug

1. Go to homework creation page
2. Create a homework with the description: "no markdown"
3. Edit the homework and change the description to "still no markdown"
4. Save edited homework.

## Current _broken_ behavior

A `markdown_used`-event is logged.

## Expected behavior
No `markdown_used`-event should be logged.

## Tested device

- App version: 1.7.9

PR #1113, opened 2023-10-05
Continue refactoring homework dialog.

Jonas-Sander:
Continue refactoring the homework dialog (see #1112)
* Add `loadCloudFiles` to `HomeworkDialogApi`
* Make attributes of `HomeworkDialogApi` private
* Add empty state and edit state widget tests (can probably be replaced by golden tests in the future)
* Add widget that inputs data and compares it to the `UserInput`

PR #1116, opened 2023-10-06
Continue refactoring homework dialog.

Jonas-Sander:
Continuation of #1113.

* Added a test for editing a homework
* Move analytics from the UI into the bloc.
* Remove `oldHomework` from `bloc.submit` because it's unnecessary.
* Change bloc method: `bool isValid()` to `void validateInputOrThrow()`.

I'm trying to not change logic in the refactorings, so that structural and behavioral changes are seperated. This is why I'm leaving this broken behavior in for now #1115.

### DIFF
--- a/app/lib/blocs/homework/homework_dialog_bloc.dart
+++ b/app/lib/blocs/homework/homework_dialog_bloc.dart
@@ -6,6 +6,7 @@
 //
 // SPDX-License-Identifier: EUPL-1.2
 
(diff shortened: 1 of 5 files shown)


Jonas-Sander:
I recommend using the split diff view, else the diff looks pretty chaotic

grafik

nilsreichardt (approved):
LGTM

PR #1120, opened 2023-10-09
Remove `Validator` from `HomeworkDialogBloc`.

Jonas-Sander:
I didn't feel like they provided that much value so I removed them.


nilsreichardt (approved):
LGTM

Issue #1117, opened 2023-10-09
Clicking "Save" on a homework multiple times will create duplicates.


Jonas-Sander:
## Describe the bug
Clicking on "Speichern" in the add homework dialog will create the homework multiple times.

## Steps to reproduce the bug
1. Go to homework page
2. Click "+" button
3. Add titel, course and due date
4. Add a large attachment to the homework (this will prevent the dialog from closing right away, so one can press "Speichern multiple times more easily)
5. Click "Speichern" multiple times

## Current _broken_ behavior
The homework will be created several times

## Expected behavior
The homework should be saved only once. The "Speichern" button should probably be disabled after pressing.

## Screenshots, videos or logs

[](https://github.com/SharezoneApp/sharezone-app/assets/29028262/b9f7916f-06fd-4884-9ba6-61b0fb8cc2bf)

## Tested device
- App version: 1.7.9

nilsreichardt:
Closed with https://github.com/SharezoneApp/sharezone-app/pull/1699

PR #1127, opened 2023-10-12
Rework `HomeworkDialogBloc` .

### LINKED ISSUE #1115: Markdown analytics are wrong when editing a homework.

Jonas-Sander:
Replaced `HomeworkDialogBloc` with a `Bloc` from [package:bloc](https://pub.dev/packages/bloc).  
Added unit and golden tests.

I tried to keep the behavior mostly the same. There are still some of the same bugs as before (e.g. #1117), they should be fixed in separate PRs.

Differences to old behavior:
* If any field is changed the quit prompt will appear (e.g. submission enabled, submission time and is private were ignored before)
* Pressing save will show errors (red text) like before, but not a `SnackBar`

Not so great:
* `PrefilledTextField` will we rebuild each time the text changes (title, description) - didn't see any problems because of this though
* Code quality is mediocre (e.g. using `HomeworkDto` for the current state in the bloc which lead to _horrible_ hacks for the todo `DateTime`)
* Quality of tests if mediocre, good amount of duplication. Also instead of testing each thing seperately I threw some different scenarios (private, not private, with submissions) together which just tested different things at the same time.
* Markdown analytics not tested automatically, just manually once
* No automatic tests for awaiting or not awaiting API calls depending on if attachments are added because of the Firestore offline behavior where Future won't complete (it works though, tested it manually several times)

Good:
* Good amount of test coverage
* Several golden tests (happy path and errors)
* UI is decoupled from business logic
* Should be a bit better to add functionality and maintain than before
* 🗿

https://github.com/SharezoneApp/sharezone-app/assets/29028262/86ef445f-2424-4607-b3fc-d403d70eafa2   

https://github.com/SharezoneApp/sharezone-app/assets/29028262/64332530-0181-4bbf-8a35-d31e62405bc8 

Fixes #1115


app/lib/pages/homework/homework_dialog.dart
```diff
@@ -174,50 +175,85 @@ class __HomeworkDialogState extends State<__HomeworkDialog> {
...
+            ScaffoldMessenger.of(context).hideCurrentSnackBar();
+            showLeftRightAdaptiveDialog(
+              content: Text(
+                  'Hausaufgabe konnte nicht gespeichert werden.\n\n$error\n\nFalls der Fehler weiterhin auftritt, kontaktiere bitte den Support.'),
+              left: null,
+              right: AdaptiveDialogAction.ok,
+              context: context,
+            );
+        }
+      },
```
Jonas-Sander, line 196: Used [`package:bloc_presentation`](https://pub.dev/packages/bloc_presentation) here.

Jonas-Sander:
@nilsreichardt Ready to review. Tbh some of the code is a bit shitty, test code is not perfect (good amount of duplication) but my motivation is decreasing and I don't want this to be open for forever. Do feel free to critique, and suggest improvements though.

nilsreichardt (approved):
image

Is that correct that the due date is red when it's already filled out? As a user, when a due date is red, I would assume that something is wrong with the due date.

Is it still possible to display still a snack bar when there is an error? I know that it's now difficult to display a concrete error message, but we could still display something like "Bitte fülle alle erforderlichen Felder aus". This provides the user a clear action what to do (and it's more a11y friendly because the snack bar will automatically be read by the screen reader when it pops up).

LGTM

Jonas-Sander:
> Is that correct that the due date is red when it's already filled out? As a user, when a due date is red, I would assume that something is wrong with the due date.

No that was a bug in the goldens test code setup that I now fixed.   

> Is it still possible to display still a snack bar when there is an error? I know that it's now difficult to display a concrete error message, but we could still display something like "Bitte fülle alle erforderlichen Felder aus". This provides the user a clear action what to do (and it's more a11y friendly because the snack bar will automatically be read by the screen reader when it pops up).
> 
> LGTM

I added a SnackBar that always says: "Bitte fülle alle erforderlichen Felder aus" if the save button is pressed and not all necessary fields are filled out 👍

PR #1130, opened 2023-10-20
Add `sz exec` command.

Jonas-Sander:
Requires #1131

The `sz exec` command runs a custom command for all / multiple packages.   
E.g. `sz exec -- fvm dart fix --apply`.

The flags `onlyDart` and `onlyFlutter` can be used to run the command only in pure Dart or only Flutter packages.  
E.g. `sz exec -onlyDart -- fvm dart pub get`.


Jonas-Sander:
@nilsreichardt Can be reviewed now

nilsreichardt (approved):
LGTM

PR #1131, opened 2023-10-20
Use `process_runner` for Sharezone CLI.

Jonas-Sander:
By using [`process_runner`](https://pub.dev/packages/process_runner) (which uses the [`process`](https://pub.dev/packages/process) package) we:

**Make testing our CLI easier in the future**.   
`process_runner` uses the `ProcessManager` class from the `process` package internally, which is used to provide a level of indirection from the static methods in the [Process](https://api.dart.dev/stable/3.1.3/dart-io/Process-class.html) class, which in turn allows the underlying implementation to be mocked out or decorated for testing and debugging purposes.

**Remove custom hacks**  
In my custom `runProcess` method I had to work around some quirks of the Dart native Process handling. `process_runner` deals with that as well, so that we have less custom code to worry about.


tools/sz_repo_cli/lib/src/commands/src/deploy_android_command.dart
```diff
@@ -217,22 +224,18 @@ class DeployAndroidCommand extends Command {
...
+      workingDirectory: Directory(
+          path.join(_repo.sharezoneFlutterApp.location.path, '/android')),
+      addedEnvironment: {
         'TRACK': track,
         'ROLLOUT': rollout,
-        // Sets the number of retries for uploading the app bundle to Google
-        // Play. This is needed because sometimes the upload fails for unknown
-        // reasons.
-        //
-        // See: https://github.com/fastlane/fastlane/issues/21507#issuecomment-1723116829
```
nilsreichardt, line 235: Wasn't the comment helpful? :(

nilsreichardt (approved):
Nice, LGTM

tools/sz_repo_cli/lib/src/commands/src/deploy_android_command.dart
Jonas-Sander: Ah that was on accident, good catch! Its helpful :)

PR #1132, opened 2023-10-20
Replace `dart:io` file code with `package:file`.

Jonas-Sander:
We use `package:file` to create a layer of indirection (`FileSystem`) so that code can be tested more easily (similar to #1131).   
Additionally I created a `Context` class, just so that one doesn't have to pass many parameters to all `Command` classes.

Requires #1130
Requires #1131


Jonas-Sander:
@nilsreichardt Can be reviewed now

nilsreichardt (approved):
LGTM

PR #1133, opened 2023-10-25
Move files

Jonas-Sander:
I moved several files around.    

Most notably I moved all files out of `app/lib/pages` and `app/lib/blocs` into their designated domain subfolder (e.g. `app/lib/pages/settings/settings_page.dart` to `app/lib/settings/settings_page`) and thus deleted `app/lib/pages` and `app/lib/blocs`.   

Beforehand we were already mix and matching both approaches (top level pages/bloc folder and domain folders), now we use mostly domain folders (except `util`, `widgets` and `main`).


nilsreichardt (approved):
LGTM

Issue #53, opened 2021-11-04
Homework for 1st november is shown as "later" instead of "tomorrow" category in homework page on 31st october.


Jonas-Sander:
## Describe the bug
A Homework that is due on the 1st november is shown in the"later" category instead in the "tomorrow" category if the current date is the 31st october.

## Steps to Reproduce
TODO: Haven't reproduced myself yet.

1. Set the date of the device to 31st october
2. Open the homework page in Sharezone
3. Add homework for 1st november
4. See that homework is in "later" category

## Current _broken_ behavior
Homework is in "later" category

## Expected behavior
Homework is in "tomorrow" category

## Screenshots, videos or logs
![image](https://user-images.githubusercontent.com/29028262/140388307-eaac40cd-0be0-4124-8633-4cc2bfacfd98.png)

See user report in Discord: [link](https://discord.com/channels/545668860859252746/616181222577668106/904473090015236177)

## Additional Context
Original issue: https://gitlab.com/codingbrain/sharezone/sharezone-app/-/issues/1193

## Tested device
Unknown (user report)

PR #1134, opened 2023-10-25
Fix wrong section title for homeworks in next month.

Jonas-Sander:
The heading "Später" (later) was wrongly shown if the current day is the end of month and the due date for the homework is the start of the next month. This was caused by using `date.addDaysWithNoChecking` instead of `date.addDays`.

Fixes https://github.com/SharezoneApp/sharezone-app/issues/53.


nilsreichardt (approved):
LGTM

PR #1147, opened 2023-11-03
Load remote config values in web right away.

### LINKED ISSUE #1108: Web should not use the "on next restart" remote config strategy.
As shown in #1070 new values are not loaded on the first start on web. Since many users will only load the web app once we need to load the values from remote config on first startup on web. This is necessary so we can e.g. change the fallback price for Sharezone Plus on web. 

The remote config behavior measured in #1070:

| Action | Has remote/updated values|
|---|---|
| First open (Incognito Tab) | ❌ |
| Closing tab and re-opening (Not leaving Incognito) | ✅ |
| Reloading page | ✅ |
| Closing Incognito completely and re-opening | ❌ |
| Logging out and logging back in with same account | ❌ |
| Logging out and logging back in with other account | ❌ |
| Closing Tab and reopen (Not leaving incognito) and login with another account | ✅ |

Jonas-Sander:
Fixes #1108

### DIFF
--- a/app/lib/main/plugin_initializations.dart
+++ b/app/lib/main/plugin_initializations.dart
@@ -83,10 +83,12 @@ class PluginInitializations {
     });
 
     try {
-      if (flavor == Flavor.dev) {
-        // Since we depend on some values from our Remote Config in the dev
-        // environment, we can't use the "Load new values for next startup"
-        // strategy.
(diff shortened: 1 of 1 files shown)


nilsreichardt (approved):
LGTM

PR #1151, opened 2023-11-04
Add dependency license check to GitHub Actions

### LINKED ISSUE #23: Add CI enforced License check of dependencies if we open-source
If we open-source our software we need to make sure that we don't use a package that is licensed under a strong copy-left license (e.g. GPLv3) that would force us (if I understood it correctly) to also make our project licensed under the same license (e.g. we would need to use GPLv3).  

This could be done with something like [dart_license_checker](https://pub.dev/packages/dart_license_checker) or [FOSSA](https://fossa.com).

Jonas-Sander:
Check that all packages we use have licenses that we are allowed / we want to use via [`package:license_checker`](https://pub.dev/packages/license_checker).

(Unfortunately every new local `lib` package will have to be added to the `licenses_config.yaml` manually.)

If it fails the output will list all packages with a non-permitted license.   
See this example where I removed the `BSD-3-Clause` license from `permittedLicenses`:
```
 ┌─────────────────────────────────────────────────────────┐
│                             Package Name  License       │
├─────────────────────────────────────────────────────────┤
│                         flutter_markdown  BSD-3-Clause  │
│                              built_value  BSD-3-Clause  │
│                               characters  BSD-3-Clause  │
│                             checked_yaml  BSD-3-Clause  │
│                          cloud_firestore  BSD-3-Clause  │
└─────────────────────────────────────────────────────────┘
```

Closes #23


Jonas-Sander:
> LGTM
> 
> > (Unfortunately every new local lib package will have to be added to the licenses_config.yaml manually.)
> 
> We can later add a command to our CLI that checks if each package we use is mentioned in `licenses_config.yaml`. This command can then be included in our GitHub checks.

This isn't necessary, the pipeline will fail if we don't add it to the file. I just meant that adding it manually still has to be done for each new package (it isn't automated).

nilsreichardt:
> This isn't necessary, the pipeline will fail if we don't add it to the file. I just meant that adding it manually still has to be done for each new package (it isn't automated).

Ah nice, this shouldn't be a big problem 👍

PR #1156, opened 2023-11-06
Update codemagic-cli-tools to 0.47.0.

Jonas-Sander:
The bug that I worked around with #1153 is fixed with 0.47.0 of codemagic-cli-tools.


nilsreichardt (approved):
LGTM

PR #1160, opened 2023-11-21
Fix homework dialog todoUntil hack.

Jonas-Sander:
Add and use `_DateSelection` class inside `HomeworkDialogBloc`. Previously I used special "not selected" values for the internal `_HomeworkDto.todoDate` that the bloc used for its state which was very hacky and won't work when adding lesson selection chips.


nilsreichardt (approved):
LGTM

Issue #1159, opened 2023-11-13
[Feature Request] Upgrade to Material Design 3 / M3


Lauritz-Tieste:
To further improve the look of the Sharezone app, an upgrade to Material Design 3 / M3 and dynamic color would be nice.

[https://m3.material.io/](https://m3.material.io/)

deepraj02:
what changes are you expecting??

Lauritz-Tieste:
Like upgraded widgets and dynamic color

nilsreichardt:
It's a bit more work and requires a few design changes. I think this ticket needs to be implemented by the Sharezone team.

PR #1164, opened 2023-11-29
Upgrade to Flutter v3.16.3

nilsreichardt:
## Description

This PR upgrades to Flutter v3.16.3. We faced the following breaking changes:

* [Deprecated just-in-time navigation pop APIs for Android Predictive Back](https://docs.flutter.dev/release/breaking-changes/android-predictive-back)
* [The ThemeData.useMaterial3 property is now set to true by default](https://docs.flutter.dev/release/breaking-changes/material-3-default)

## Material 3

For now, we just opt out of Material 3 because it requires some more manual migration and isn't our current focus (#1159).

## Predictive Back

> Android 14 introduced the [Predictive Back feature](https://developer.android.com/guide/navigation/predictive-back-gesture), which allows the user to peek behind the current route during a valid back gesture and decide whether to continue back or to cancel the gesture. This was incompatible with Flutter’s navigation APIs that allow the developer to cancel a back gesture after it is received.
>
> With predictive back, the back animation begins immediately when the user initiates the gesture and before it has been committed. There is no opportunity for the Flutter app to decide whether it’s allowed to happen at that time. It must be known ahead of time.
>
> For this reason, all APIs that allow a Flutter app developer to cancel a back navigation at the time that a back gesture is received are now deprecated. They have been replaced with equivalent APIs that maintain a boolean state at all times that dictates whether or not back navigation is possible. When it is, the predictive back animation happens as usual. Otherwise, navigation is stopped. In both cases, the app developer is informed that a back was attempted and whether it was successful.

From https://docs.flutter.dev/release/breaking-changes/android-predictive-back#background

We mostly have used `WillPopScope` for confirmation dialogs, so the migration looked like this:

```dart
WillPopScope(
  onWillPop: () async {
    final bool? shouldPop = await _showBackDialog();
    return shouldPop ?? false;
  },
  child: child,
)
```

to

```dart
return PopScope(
  canPop: false,
  onPopInvoked: (bool didPop) async {
    if (didPop) {
      return;
    }
    final NavigatorState navigator = Navigator.of(context);
    final bool? shouldPop = await _showBackDialog();
    if (shouldPop ?? false) {
      navigator.pop();
    }
  },
  child: child,
)
```

https://docs.flutter.dev/release/breaking-changes/android-predictive-back#migrating-a-back-confirmation-dialog

### Run the application with predictive back

1. Run Android 33 or above.
2. Enable the feature flag for predictive back on the device under “Developer options”. This will be unnecessary on future versions of Android.
3. Set `android:enableOnBackInvokedCallback="true"` in android/app/src/main/AndroidManifest.xml. If needed, refer to [Android’s full guide](https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture) for migrating Android apps to support predictive back.
4. Make sure you’re using version 3.14.0-7.0.pre of Flutter or greater.
5. Run the app and perform a back gesture (swipe from the left side of the screen).

https://docs.flutter.dev/release/breaking-changes/android-predictive-back#supporting-predictive-back

Unfortunately, I wasn't able to get predictive back animations running in a Flutter app. They work for me in the native settings app, but not in a Flutter app (even with the example app). I think it shouldn't be that a big deal because it should work that way as we have it. Here is a demo of how it should look: https://github.com/flutter/flutter/pull/120385.

## Testing

I tested the changes manually.


Sharezone-Admin:
⬇️ Builds generated by [Codemagic](https://codemagic.io) for commit c7b888afeaa701e495a0964fabeff7aec0ad626b ⬇️

| iOS | | Android |
|:-:|:-:|:-:|
| iOS QR code <br /> [Download link](https://api.codemagic.io/artifacts/.eJwVyktygjAAANC7uHeGT0Jl4QKUn9EEA6Jk40SLFIhAgSbi6du-9VuUzj9X3LVUQUyjtNODyE7lEu965cgKaD9mQtmTaiVpjaRX1J_foCGPMbrYmAdMjw14e3EJ6thAptpdw_Fs3IcqBnbXIwo-EBtkNHE0OLrl-j2k4_ucFU9B6GmEvC1xFKpLi_P6GBOq7R8ZmXwLnMzthEJLAYVm1gnJcVZGn3IOdG9KmJjh3BzgMzgH35nzKiAjUeXy3EZex2G9OVobB-JwlX_5h13CVuktMLvcY2-ebavJa_8yrod9lbmFb8Zi-bAT9yqboukGM8n2qJ_lgdxF3whLrdeLX6vEYbM.xnPq8pLwqiD3wHkvFzfyOnmsNug) | | Android QR code <br /> [Download link](https://api.codemagic.io/artifacts/.eJwVwclygjAAANB_8c4MU5Hl4CFKWCRWG7bihYFEacACjYRAvt7pe5sG_Ds8iZ5IKzCdmNoFaqNwNvOr9lQkHNZOh-nHH80NryMV3uJpKgffsMq02U5tgxkT9QHycAml3QeMegjuhByihdn3aRado8wu_4zDsm8vx_SsAYCOxdlTCH6rkRikmUrM-oO8uAZkv6faTYvslhU-uDA3QnxA1hJXag30L5g9VBogkIoY81ynLb71cLDUKls13AE3xaw9tetcaWJHE2_8ueIzd181lVHAa0cY_ohP1auY7XirKIv8vG3QyDKVgNspsWqeKIc84LIuDyJXXhZLKRq532_eZu1l1w.hs6NzhQN2c23-SOY6pzNaH8I5eM) |

<sub>(expires Sun, 7 Jan 2024 16:31:44 GMT)<br/>Codemagic build: [657341ee075ffd89e80f359a](https://codemagic.io/app/629c82ea463af7ff553fc7a5/build/657341ee075ffd89e80f359a); comment generated with [Codemagic App Preview](https://github.com/nilsreichardt/codemagic-app-preview)</sub>

PR #1171, opened 2023-12-07
Pass `successUrl` and `cancelUrl` to `createStripeCheckoutSession` cloud function

nilsreichardt:
## Demo

https://github.com/SharezoneApp/sharezone-app/assets/24459435/723d85a3-55ce-485a-95fa-f9f1edd05c53

### DIFF
--- a/app/lib/sharezone_plus/page/sharezone_plus_page_controller.dart
+++ b/app/lib/sharezone_plus/page/sharezone_plus_page_controller.dart
@@ -79,7 +79,10 @@ class SharezonePlusPageController extends ChangeNotifier {
   }
 
   Future<void> _buyOnWeb() async {
-    final webAppUrl = _getBaseWebAppUrl();
+    // The URL is used to redirect the user back to the web app after the
+    // payment is completed or canceled. We only use the base URL because we a
+    // whitelist for the allowed URLs.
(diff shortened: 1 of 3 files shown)


app/lib/sharezone_plus/page/sharezone_plus_page_controller.dart
```diff
@@ -79,8 +79,19 @@ class SharezonePlusPageController extends ChangeNotifier {
   }
 
   Future<void> _buyOnWeb() async {
+    // The URL is used to redirect the user back to the web app after the
+    // payment is completed or canceled. We only use the base URL because we a
+    // whitelist for the allowed URLs.
```
Jonas-Sander, line 84: Last sentence is incomplete. Also I don't follow the argument, I mean the backend could also just match the base url if a longer url is send, no? I guess the argument is so that we have to do less parsing in the backend we only send the base url, correct?

Jonas-Sander (approved):

app/lib/sharezone_plus/page/sharezone_plus_page_controller.dart
nilsreichardt: Changed the backend that it only needs to match the beginning of the URL ✅

PR #1222, opened 2024-01-05
Add dev environment to `sz deploy web-app` command

nilsreichardt:
This PR adds the options to deploy our web app to our dev environment, e.g.

```sh
sz deploy web-app \
  --stage alpha \
  --flavor dev
```

### DIFF
--- a/tools/sz_repo_cli/lib/src/commands/src/deploy_web_app_command.dart
+++ b/tools/sz_repo_cli/lib/src/commands/src/deploy_web_app_command.dart
@@ -59,7 +59,7 @@ class DeployWebAppCommand extends CommandBase {
       ..addOption(
         flavorOptionName,
         allowed: _webFlavors,
-        help: 'The flavor to build for. Only the "prod" flavor is supported.',
+        help: 'The flavor to build for.',
         defaultsTo: 'prod',
       );
   }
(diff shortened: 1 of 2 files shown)


tools/sz_repo_cli/lib/src/commands/src/deploy_web_app_command.dart
```diff
@@ -192,27 +190,3 @@ class DeployWebAppCommand extends CommandBase {
     return currentCommit;
   }
 }
-
-class _WebAppConfig {
```
nilsreichardt, line 196: I remove the `_WebConfigConfig` class since it was too overkill, `_webAppConfigs` wasn’t suitable to support multiple projects with the same stage and `_WebConfigConfig` would duplicate the Firebase Hosting targets.

tools/sz_repo_cli/lib/src/commands/src/deploy_web_app_command.dart
```diff
@@ -46,18 +55,20 @@ class DeployWebAppCommand extends CommandBase {
         googleApplicationCredentialsOptionName,
         help:
             'Path to location of credentials .json file used to authenticate deployment. Should only be used for CI/CD, developers should use "firebase login" instead.',
+      )
+      ..addOption(
+        flavorOptionName,
+        allowed: _webFlavors,
+        help: 'The flavor to build for. Only the "prod" flavor is supported.',
```
Jonas-Sander, line 62: This is outdated

Jonas-Sander (approved):
Nice 👍

PR #1233, opened 2024-01-06
Move widgets of Sharezone Plus page into it's own package

nilsreichardt:
This PR moves the UI widgets of the Sharezone Plus page into its own package called `sharezone_plus_page_ui`. The motivation behind this is to reuse these widgets on our website. This package only contains the UI not any logic, like `SharezonePlusPageController` because the logic is different between our app and the website (in our app you have an account, on our website you don't have an account).


lib/sharezone_plus/sharezone_plus_page_ui/lib/src/call_to_action_button.dart
```diff
@@ -0,0 +1,52 @@
...
+//
+// You may obtain a copy of the Licence at:
+// https://joinup.ec.europa.eu/software/page/eupl
+//
+// SPDX-License-Identifier: EUPL-1.2
+
+import 'package:flutter/material.dart';
+import 'package:sharezone_widgets/sharezone_widgets.dart';
+
+class CallToActionButton extends StatelessWidget {
```
nilsreichardt, line 12: Should we move this to `sharezone_widgets` or keep it in `sharezone_plus_page_ui` for now?

Jonas-Sander: Not sure, just do what you think is the best

Jonas-Sander (approved):

lib/sharezone_plus/sharezone_plus_page_ui/lib/src/call_to_action_button.dart
nilsreichardt: I would keep it in `sharezone_plus_page_ui`. If we need anywhere else, we can move it to `sharezone_widgets`.

PR #1257, opened 2024-01-14
Remove "Rate app" section in theme settings.

Jonas-Sander:
| Before | After |
|--------|--------|
| before | after |

I think we added this way back when we added/unlocked dark mode for everyone

### DIFF
--- a/app/lib/settings/src/subpages/theme/theme_page.dart
+++ b/app/lib/settings/src/subpages/theme/theme_page.dart
@@ -6,7 +6,6 @@
 //
 // SPDX-License-Identifier: EUPL-1.2
 
-import 'package:app_review/app_review.dart';
 import 'package:bloc_provider/bloc_provider.dart';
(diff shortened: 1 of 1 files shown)


nilsreichardt:
> I think we added this way back when we added/unlocked dark mode for everyone

We added it there because it's one of the rare spots where a user has hopefully time to review us (compared to showing a popup when starting the app -> users want to add a homework and dismiss the popup). And hopefully has a positive vibe when their see there is dark / light mode and motivates the user to rate it.

Unfortunately, we don't have analytics data for this button.

But if you like, feel free to remove this card.

nilsreichardt (approved):

Jonas-Sander:
Idk, I'm also fine with leaving it. I think it will be very rare though that a user will review it through that dialog. I think we should rather ask for a rating after checking a homework as completed for example. On the other hand this dialog shouldn't hurt us there so.... let's just leave it for now.

PR #1258, opened 2024-01-14
Increase border width of `CustomCard` by 0.5

nilsreichardt:
| Before | After |
|--------|--------|
| Screenshot 2024-01-14 at 22 54 52 | Screenshot 2024-01-14 at 22 54 33 |

I think this looker better because with border width 1 the rendering of the borders is not so clear.


Jonas-Sander (approved):
Yeah, looks way better

PR #1270, opened 2024-01-19
Add weekday in date picker value

nilsreichardt:
Most students think in weekdays and not in dates.


Jonas-Sander:
Shouldn't the date/language be German in the goldens?

Jonas-Sander (approved):

PR #1274, opened 2024-01-19
Reduce text size of homework tile

nilsreichardt:
| Before | After |
|--------|--------|
| Screenshot 2024-01-19 at 14 29 18 | image | 

Homework tiles used too much space. I think this is better.


Jonas-Sander (approved):
I kinda liked having bigger tiles with more spacing. Ofc text size was not fitting with the rest. Switching between both the smaller tiles seem so squished. (I still want to make this change, just random thoughts here)

PR #1281, opened 2024-01-29
Add `createdOn` for calendrical events

### LINKED ISSUE #1280: Add `createdAt` field to events
Currently, events don't have a `createdAt` field. However, it would be helpful they would have one.

nilsreichardt:
Closes #1280 

## Summary by CodeRabbit

- **New Features**
	- Introduced a "Created On" date for calendrical events, enhancing event tracking and organization.
- **Refactor**
	- Improved the handling of event creation times across the app for consistency.

### DIFF
--- a/app/lib/calendrical_events/models/calendrical_event.dart
+++ b/app/lib/calendrical_events/models/calendrical_event.dart
@@ -6,6 +6,7 @@
 //
 // SPDX-License-Identifier: EUPL-1.2
 
+import 'package:cloud_firestore_helper/cloud_firestore_helper.dart';
 import 'package:date/date.dart';
 import 'package:group_domain_models/group_domain_models.dart';
 import 'package:sharezone/timetable/src/models/lesson.dart';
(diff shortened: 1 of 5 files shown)


app/lib/calendrical_events/models/calendrical_event.dart
```diff
@@ -15,6 +16,13 @@ import 'package:time/time.dart';
 import 'calendrical_event_types.dart';
 
 class CalendricalEvent {
+  /// The date and time when the event was created.
+  ///
+  /// Clients with a lower version than 1.7.9 will not have this field.
+  /// Therefore, we will always have events without a [createdOn] field in the
+  /// database.
+  final DateTime? createdOn;
```
nilsreichardt, line 24: Technically, we don't need to parse the `createdOn` field because we don't need in the client. Should we still add it to the DTO and parse it?

Jonas-Sander: I'm fine with it 👍

Jonas-Sander (approved):

PR #1288, opened 2024-02-03
Upgrade CI/CD pipelines from `macos-13` to `macos-14`

nilsreichardt:
`macos-14` runner are using M1 CPUs: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories.

This change also allows us to run our Flutter golden tests with GitHub Actions instead of Codemagic.


nilsreichardt:
New idea: keep old macos runner, only use macos-14 for golden tests

nilsreichardt:
Closing because of #1325

PR #1290, opened 2024-02-06
Use `sz build web` in web-app preview job and add `preview` stage to it.

Jonas-Sander:
Use `sz build web` in web-app preview job and add `preview` stage to it.

Together with #1291 this will display a "PREVIEW" banner for all previews.

Since this uses `pull_request_target` one can't see it in the preview here. I checked manually though (by overriding `pull_request_target` with `pull_request`)


nilsreichardt (approved):
LGTM

PR #1292, opened 2024-02-07
Add `--only` option to sharezone cli.

Jonas-Sander:
The option allows to run a `ConcurrentCommand` only for the specified package names.

E.g.: `sz analyze --only sz_repo_cli`, `sz test sharezone,abgabe_client_lib`.

## Summary by CodeRabbit

- **New Features**
	- Introduced an option to execute tasks for specific packages only, enhancing task management efficiency.

PR #1293, opened 2024-02-07
Use `super.packagesToProcess` so that `ConcurrentCommand` filters are automatically applied.

Jonas-Sander:
Use `super.packagesToProcess` so that #1292 also applies to `sz test`


nilsreichardt (approved):
LGTM

PR #1311, opened 2024-02-23
Fix setting submission to null when submission is disabled

### LINKED ISSUE #1306: Former submission homeworks have the wrong subtitle in the dashboard view
Once "Mit Abgabe" is activated, the due to time is set to 23:59 or something else. When the "Mit Abgabe" is now deactivated again, the due to time is still 23:59. This causes the problem, that the due to date is displayed in the dashboard instead of "bis Morgen"

Example:

EvilMonkey09:
When the "Abgabe" was activated, the submissionTime was set to 23:59. If you deactivated the "Abgabe" the submissionTime wasn't set to null, because of the _dateSelection.copyWith which ignores null.
This caused the problem, that in the Dashboard the dueDate was displayed in red and not "Bis morgen!".
[Issue #1306](https://github.com/SharezoneApp/sharezone-app/issues/1306)
306628701-b161437d-40ad-4f50-bd58-00689892e194

Closes #1306 

## Summary by CodeRabbit

- **Refactor**
	- Updated the logic for date selection handling in the homework dialog to improve user interaction.


Jonas-Sander:
Hi, thanks! :)
Could you try and make a regression test for that?
Look at the test at line 330 in `app\test\homework\homework_dialog_bloc_test.dart` to see how that should work.  

I guess the test should check that if one enables and disables the submissions for a homework and then saves the homework, that the `UserInput` passed to the `MockHomeworkDialogApi` (`homeworkDialogApi.userInputToBeCreated`) will have a DateTime with zeros for hours, minutes, seconds etc.

nilsreichardt (approved):
LGTM

nilsreichardt:
@Jonas-Sander I added a small test. Could you also review?

Jonas-Sander (approved):

PR #1318, opened 2024-02-26
Randomize tests; add `--test-randomize-ordering-seed` option to `sz test`.

Jonas-Sander:
By default Dart/Flutter always run tests in the same order. This can be changed by specifying `--test-randomize-ordering-seed` flag. We added this flag to `sz test` and also use the value "random" by default.   

If a test fails because of the randomization one can see that seed at the very start of the output of the failing test. This seed can then be passed like this: `sz test --test-randomize-ordering-seed 1784202788`.

Requires #1319 

## Summary by CodeRabbit

- **New Features**
	- Added an option to specify a seed for randomizing test case execution order.
	- Users can now provide a 32-bit unsigned integer or "random" as the seed.
	- Improves test robustness by allowing control over test case execution order.


Jonas-Sander:
Test is failing because #1319 is not merged yet.

nilsreichardt (approved):
LGTM

PR #1319, opened 2024-02-26
Fix flaky test using mockito by using `provideDummy`.

Jonas-Sander:
This test was failing when using #1318 with the ` --test-randomize-ordering-seed 1`.

Required for #1318 

## Summary by CodeRabbit

- **Tests**
	- Updated the setup for `Past Calendrical Events Page` tests to use a dummy implementation for state testing due to limitations with mocking frameworks.


nilsreichardt (approved):
LGTM

PR #1349, opened 2024-03-01
Add option to change the type of user

### LINKED ISSUE #556: User should be able to change the user type without contacting the support
## Description

Currently, users need to contact the support to change their user type. It would be great, if the user can change their user type without contacting the support, e.g. just with a button in the app.

When the user changes the user type, we need to show few information:
* The changes are applied after restarting the app
* If the user has a subscription, the user needs to cancel the subscription and subscribe again (because every user type has a separate subscription)

nilsreichardt:
## Description

A user can change twice every two weeks their type of user in the app setting without contacting the support.

### Changing type of user

https://github.com/SharezoneApp/sharezone-app/assets/24459435/9300571b-5e09-4b96-b103-76b41b7cfec2

### When the user changed too often the type of user

https://github.com/SharezoneApp/sharezone-app/assets/24459435/cce72970-4b1b-4aba-926e-efca5760554d

## Related tickets

Closes #556


app/lib/settings/src/subpages/my_profile/change_type_of_user/change_type_of_user_repository.dart
```diff
@@ -0,0 +1,25 @@
...
+//
+// You may obtain a copy of the Licence at:
+// https://joinup.ec.europa.eu/software/page/eupl
+//
+// SPDX-License-Identifier: EUPL-1.2
+
+import 'package:cloud_functions/cloud_functions.dart';
+import 'package:user/user.dart';
+
+class ChangeTypeOfUserRepository {
```
Jonas-Sander, line 12: Its not really a Repository, lets just call it `ChangeTypeOfUserService` instead I'd say.

Jonas-Sander (approved):

app/lib/settings/src/subpages/my_profile/change_type_of_user/change_type_of_user_repository.dart
Jonas-Sander: Or anything else, just not Repository since it does not follow the Repository pattern.

Jonas-Sander:
Don't forget to change the PR if you implement my suggestion on the backend (`errorId`)

Jonas-Sander (approved):

PR #1357, opened 2024-03-03
Migrate to FVM 3

nilsreichardt:
- Migrate to FVM 3
- Use Flutter FVM Config Action 2.0
- Reduces CI setup for Flutter from 60s to 30s

Note: Every developer needs to upgrade to FVM 3!


nilsreichardt:
Some workflows are failing because of `pull_request_target`. No problem, when this PR is merged.

Jonas-Sander (approved):

PR #1360, opened 2024-03-04
Add prefix `app` to `build` and `deploy` app commands

nilsreichardt:
Since we are adding more and more different `build` and `deploy` commands (like `sz deploy website`), I added a `app` prefix for the app commands:

From `sz deploy ios` to `sz deploy app ios` (same applies for `macos`, `android` and `web`).

### DIFF
--- a/.github/workflows/alpha.yml
+++ b/.github/workflows/alpha.yml
@@ -252,7 +252,7 @@ jobs:
           # parsing in the shell, potentially leading to incorrect handling of
           # the multiline string argument.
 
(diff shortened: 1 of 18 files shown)


Jonas-Sander:
I think I'd prefer `sz build app ios`. WDYT?

nilsreichardt:
Build fails because of `pull_request_target`

Jonas-Sander (approved):
Nice, thanks! :)

PR #1362, opened 2024-03-04
Fix deploy commands to use new build command

nilsreichardt:


Jonas-Sander (approved):
Maybe we should think about the pros/cons of just using the class/functions directly in Dart instead of calling the `sz ...` for the future. (Merge this anyways, just as a random thought)

PR #1378, opened 2024-03-06
Update banner in `README.md`

nilsreichardt:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-![Sharezone Banner with two phones](https://user-images.githubusercontent.com/29028262/151260826-4d13664c-8269-442c-bf78-1197899afffb.png)
+![Sharezone wallpaper](https://github.com/SharezoneApp/sharezone-app/assets/24459435/7f2d9dcc-82d7-4342-a7e2-1e10491e4a94)
 
 ## Download Sharezone
 


nilsreichardt:
@Jonas-Sander WDYT?

Jonas-Sander:
I think the blue of the current banner is nicer than a grey, boring background. I can see why you would add a subtitle with an intro to Sharezone, I'm open for that. I think the phones laying down and being "flat" is a bit more "uninteresting" visually, but if you like it better I'm open to changing it.

nilsreichardt:
The idea was to show more of the app because the current banner only shows two screens and when new people come to our GitHub Repository (in most cases they are not Sharezone users but interested in the Repository / in us) they only have this huge repository but they don't know how the app looks like (and no one is opening the web app and creates an account to be in an empty app). At first, I only used the wallpaper with the screens on the left, but this looked the image takes the entire screen :/ I think the best case is to have a 30 sec. trailer / demo video but this is much work right now.

Jonas-Sander:
I don't see how people that don't know what we're doing would be able to make real changes in our app anyways. Can't we just link to our homepage as a start? There we have the different functionality shown in a nicer format.

Jonas-Sander:
I get what you're saying though

nilsreichardt:
Closing it for now

PR #1381, opened 2024-03-08
Show created on timestamp in file sheet

nilsreichardt:
image


Jonas-Sander (approved):
I think formatting it like this "22.03.24 16:39" (or even without time) is more fitting. I wouldn't care what weekday it is (or which exact time probably - im fine with keeping it there though)

PR #1385, opened 2024-03-09
Fix color of continue button in onboarding

nilsreichardt:
| Before | After |
|--------|--------|
| Screenshot 2024-03-09 at 17 56 48 | Screenshot 2024-03-09 at 17 56 14 |

### DIFF
--- a/app/lib/auth/login_button.dart
+++ b/app/lib/auth/login_button.dart
@@ -24,31 +24,14 @@ class ContinueRoundButton extends StatelessWidget {
       button: true,
       onTap: onTap,
       label: tooltip,
-      child: IconButton(
-        tooltip: tooltip,
-        icon: Stack(
-          alignment: Alignment.center,
-          children: <Widget>[
-            AnimatedSwitcher(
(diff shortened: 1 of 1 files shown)


app/lib/auth/login_button.dart
```diff
@@ -24,31 +24,14 @@ class ContinueRoundButton extends StatelessWidget {
...
-            ),
-            const Icon(
-              Icons.chevron_right,
-              color: Colors.white,
-            )
-          ],
+      child: AnimatedSwitcher(
+        duration: const Duration(milliseconds: 250),
+        child: IconButton.filled(
+          key: ValueKey(onTap == null),
```
Jonas-Sander, line 30: This is so that it rebuilds if onTap changes right?

Jonas-Sander (approved):

app/lib/auth/login_button.dart
nilsreichardt: Yes, otherwise `AnimatedSwitcher` doesn't it needs to animate

PR #1388, opened 2024-03-09
Fix settings page for dark mode

nilsreichardt:
image

Test is in #1389 

The issue was that `Theme.of(context).listTileTheme.iconColor` was `null` for the dark mode.


Jonas-Sander (approved):
Can you post what the exact crash / error was? Is there no fallback iconColor when none is specified?

nilsreichardt:
Added it to the pull request description ✅

PR #1392, opened 2024-03-09
Add feature of iCal links

nilsreichardt:
## Description

With the iCal link feature, users can integrate their Sharezone calendar into other calendar applications, such as Google Calendar, Apple Calendar, etc. Currently, users can choose from the following sources for iCal links

* Exams
* Meetings

Later we will add support for:

* Lessons

Also, editing a link will not be possible at first.

## How does this feature work?

A user creates a Firestore document with the link information (data, sources). A cloud function pulls all relevant data from the database and puts it into an iCal format. Cloud functions take care of updating the iCal format when items (exams, meetings) are updated.

A user requests an authenticated URL from a cloud function, which can then be integrated into other applications.

## To-Do

- [ ] Add confirmation dialog for deletion
- [ ] Put feature behind feature flag
- [ ] Add as advantage to Sharezone Plus page

## View Demo

https://github.com/SharezoneApp/sharezone-app/assets/24459435/71329f3b-cce4-4b3c-8199-79063902fe82


nilsreichardt:
Backend isn't fully ready for review yet but frontend should be ready

Jonas-Sander (approved):

PR #1393, opened 2024-03-10
Refactor course create page

nilsreichardt:
In this PR, I removed code duplication (removed a lot of code) between create source without school class and with school class. Other changes:

* New: The course templates are no longer sorted by categories. When I was in school, I remember that it always took me ~10 seconds to search for a course in the course template page because there wasn't any helpful structure. So I just sorted the courses by alphabet.
* New: When opening the create course page, we should get a warning that these courses are not connected to the school class. For a better UX I added a card to select the school class in case the user is admin in one or more school classes.
* New: The confirmation bar for school class courses also shows the name of the course.
* New: The confirmation snack bar for school class courses also has an "Undo" button to delete the created course.
* New: Loading animation in the course tile when deleting a course from the snack bar.
* New: When already created a course with the name of a template and clicks on it, a confirmation dialog will be displayed.
* New: Add Ethik as subject to the templates.
* Fixed: When editing the course template and creating the source, the created course wasn't marked as created in the list.
* Fixed: The padding of the loading circle wasn't correct.
* Changed: Changed the "Hinzufügen" tooltip of the edit icon button to "Bearbeiten".

https://github.com/SharezoneApp/sharezone-app/assets/24459435/6e79713b-1d5e-4dde-90b3-0ad8f2fd4ab9

Demo Update (14.03.2024):

https://github.com/SharezoneApp/sharezone-app/assets/24459435/6de3968b-d144-4b4b-b886-49b24894fc79


Jonas-Sander:
Very nice! In your preview video you first created a "Chemie" course while "Mit keiner Schulklasse verknüpfen" was selected, then you selected the class "1b" so that it should be "verknüpft"/connected to a new course and then you created the "Deutsch" course.

So would now "Chemie" and "Deutsch" be connected to the school class "1b" or only the "Deutsch" course (i.e. new courses that were created after selecting the class)?  
I think that users could reasonably come to either conclusion, my guess would be the latter but I'm also not 100% sure.
For example one could think that the class selection would only come into effect when pressing "Fertig", so that any course created in the dialog would either be connected to a school class or not.  

In my opinion we should add some indicator to a course tile that is shown when the course is connected to a school class so that the user can be certain about the behavior. I'm not sure how though. Either some icon or maybe just the text: "Der Kurs ist mit der Klasse 1b verknüpft"? And what would be if one created two "Deutsch" courses, one connected to a school class and the other one not?

app/test_goldens/groups/src/pages/course/create/course_template_page_test.dart
```diff
@@ -0,0 +1,71 @@
...
+
+    setUp(() {
+      courseCreateBlocFactory = MockCourseCreateBlocFactory();
+      courseCreateBloc = MockCourseCreateBloc();
+
+      final random = Random(42);
+      when(courseCreateBlocFactory.create(schoolClassId: null))
+          .thenReturn(courseCreateBloc);
+      when(courseCreateBloc.isCourseTemplateAlreadyAdded(any))
+          .thenAnswer((_) => random.nextBool());
```
Jonas-Sander, line 44: I'm not sure, since you hardcoded the seed, but could this make any trouble in a golden test? Shouldn't this be hardcoded value (true/false) in the test itself?

Jonas-Sander (commented):
Why did you create `MyAdminSchoolClassesProvider` and not just included that logic in the `CourseCreateBloc`? I'm not sure if splitting the logic that way is benefical

app/test_goldens/groups/src/pages/course/create/course_template_page_test.dart
nilsreichardt: Moved it to the `CourseCreateBloc` 👍

nilsreichardt:
> Why did you create `MyAdminSchoolClassesProvider` and not just included that logic in the `CourseCreateBloc`? I'm not sure if splitting the logic that way is benefical

Moved it to the `CourseCreateBloc` 👍

app/test_goldens/groups/src/pages/course/create/course_template_page_test.dart
nilsreichardt: This shouldn't be a problem because the idea is more have some course that show the check (currently, we don't care which courses these are because the golden test doesn't test the logic of it). Since the seed is hard-coded, it will always produce the same results

nilsreichardt:
> So would now "Chemie" and "Deutsch" be connected to the school class "1b" or only the "Deutsch" course (i.e. new courses that were created after selecting the class)?

I added a snack bar that shows the following message:
If you select "Mit keiner Schulklasse verknüpfen": "Kurse, die ab jetzt erstellt werden, werden mit keiner Schulklasse verknüpft."
If you select a school class: "Kurse, die ab jetzt erstellt werden, werden mit der Schulklasse '$name' verknüpft."

I hope this clearfies it.

> In my opinion we should add some indicator to a course tile that is shown when the course is connected to a school class so that the user can be certain about the behavior. I'm not sure how though. Either some icon or maybe just the text: "Der Kurs ist mit der Klasse 1b verknüpft"? And what would be if one created two "Deutsch" courses, one connected to a school class and the other one not?

Yes, is also something I thought of. I think we can just add a subtitle to the ListTile (for course that have been already created) saying "Bereits in '10A' vorhanden". Something like this (prototype):

image

I would also change the confirmation snackbar from "Kurs 'Englisch' erstellt" to "Kurs 'Englisch' in '10A' erstellt" or "Kurs 'Englisch' ohne Verknüpfung erstellt"

However, I would do this in another PR since PR also really big and my initial only was to remove the subject categories.

Jonas-Sander (approved):

PR #1410, opened 2024-03-15
Add `--only` parameter to `website` CI

nilsreichardt:


Jonas-Sander (approved):
`website` uses packages from `lib`, wouldn't this cause them to not be analyzed?`

Issue #1312, opened 2024-02-23
Grade Calculation and overview


Arbee4ever:
Add a way to nicely add grades to individual groups, and allow the calculation of the overall year grade by using custom weights for different test types.

nilsreichardt:
Hi @Arbee4ever, thanks for your future request! We're definitely considering adding a grade feature. However, can't say if it will come and if yes, when.

PR #1414, opened 2024-03-15
Add term details page

nilsreichardt:
image

Part of #1312


Jonas-Sander (approved):
Nice 👍

Jonas-Sander:
I feel like the "Berechnung der Fachnote"/"Berechnung des Schnitts" takes a bit too much space or makes it feel kinda cluttered. Other solutions might not be that obvious for the user though.

nilsreichardt:
> I feel like the "Berechnung der Fachnote"/"Berechnung des Schnitts" takes a bit too much space or makes it feel kinda cluttered. Other solutions might not be that obvious for the user though.

Yes, this true, maybe we find later a better way (like only an edit icon in the tile)

PR #1418, opened 2024-03-16
Add grades dialog (without any logic)

nilsreichardt:
image


Jonas-Sander:
I think instead of "Thema" it should be "Titel". I personally liked "in den Schnitt mit einbringen". I also don't understand why the toggle is at optional since it always has a state that is always used. (I'm open to leaving it like like this, what I said is just my personal taste)

nilsreichardt:
I also liked "in den Schnitt mit einbringen" more but it's a bit confusing because the grade will be used to calculate the grade of the subject and everywhere else in the app we will use the word "Fachnote"

PR #1424, opened 2024-03-18
Use native browser viewer for PDF files

nilsreichardt:
https://github.com/SharezoneApp/sharezone-app/assets/24459435/c72d997e-7aea-45b0-a270-b1cee3d8cc18

Using the browser viewer for PDFs brings finally a good experience for the user:

* Great performance
* (Download PDF)
* Search in PDF
* Select text in PDF
* Print PDF
* Jump to pages
* Clickable hyperlinks

Closes #102

PR #1426, opened 2024-03-18
Remove `pdfx` from `index.html` and web

nilsreichardt:
Since #1424 we don't need `pdfx` anymore for our web platform. Therefore, we can remove the script for `pdf.js` and the exception for `cdn.jsdelivr.net` in the firewall docs.


Jonas-Sander (approved):

nilsreichardt:
We can't remove pdf.js since pdfx checks if pdf.js installed. We can reopen this PR once we replaced pdfx on Android.

PR #1435, opened 2024-03-21
Only use single random seed when using `sz test`.

Jonas-Sander:
Currently when running `sz test` without the `--test-randomize-ordering-seed` option a new seed is generated for every package. This makes it hard to know which seed was used for which package when testing several packages concurrently. WIth this PR only a single random seed for all packages being tested is generated at the beginning when running `sz test`.


nilsreichardt (approved):
LGTM

PR #1436, opened 2024-03-26
Add adding grades to a term and enhance grade logic.

Jonas-Sander:
This PR enhances the grade feature logic and adds a very hacky version of a functioning add grade dialog.  

Currently only two grading systems are implemented: 1-6 (+-) and 0-15 points, both have predefined possible values.  
In the future I'll need to make also non-discrete grading systems work (e.g. 1-6 with decimals).
In the add grade dialog only 1-6(+-) will work currently though, because the term is always created with this grade system via the UI.

Also adding real courses does not work yet, I'm just using a fake subject.  
I wasn't sure what our exact plan was like again (courses with old courses? when does stuff update? how are the subjects connected to a course?) and also I thought its maybe better to do it in a future PR.

[grades.webm](https://github.com/SharezoneApp/sharezone-app/assets/29028262/f6e40f06-5a71-491c-b300-71f474e72a88)


app/lib/grades/grades_service/src/grading_system.dart
```diff
@@ -0,0 +1,161 @@
...
+    if (res != null) {
+      return res;
+    }
+    return grade.toStringAsFixed(2).replaceAll('.', ',').substring(0, 3);
+  }
+
+  String? getDisplayableGradeIfExactMatch(num grade);
+}
+
+class OneToFiveteenPointsGradingSystem extends _GradingSystem
```
nilsreichardt, line 59: Isn't it `ZeroToFiveteen`?

nilsreichardt (approved):
LGTM

app/lib/grades/grades_service/src/grading_system.dart
Jonas-Sander: Whoops, thanks :)

PR #1440, opened 2024-04-02
Add devenv / Nix package management to repo

Jonas-Sander:
[Devenv](devenv.sh) (in combination with [direnv](https://direnv.net/)) allows developers to automatically and deterministically fetch required dependencies that are used in the project.  
The language / package management used is called [Nix](https://nixos.org/).

To use it [install devenv](https://devenv.sh/getting-started/) and [install direnv](https://direnv.net/).  
By switching into the sharezone-app directory one will automatically be dropped into a shell. This shell will have all the packages, environment variables etc installed that are specified in `devenv.nix`.   

Note: Building it the first time took a very long time because of  "androidenv.composeAndroidPackages". This is only half working, flutter doctor shows ✅ on Android but one can't run an actual emulator yet. Haven't tried building an apk. Its a WIP that I just want to submit in the currently "hopefully good enough" state. Running web via chrome did work though.  

Right now I can't really recommend it for a system where flutter etc is already set up and works. Or just try it out for lulz, idk 🤷‍♂️

Note: Nix natively support Linux and MacOS, Windows is (I think) only partially supported by using [WSL](https://learn.microsoft.com/en-us/windows/wsl/install). I'm using Linux right now.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,3 @@
+source_url "https://raw.githubusercontent.com/cachix/devenv/95f329d49a8a5289d31e0982652f7058a189bfca/direnvrc" "sha256-d+8cBpDfDBj41inrADaJt+bDWhOktwslgoP5YiGJ1v0="
+
+use devenv
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,12 @@ coverage/
 #
 # Source: https://fvm.app/documentation/getting-started/configuration
 .fvm/
+# Devenv
+.devenv*
+devenv.local.nix
+
+# direnv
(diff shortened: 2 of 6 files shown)


Jonas-Sander:
Since I need this to further develop on linux, I'm gonna go ahead and merge this without your approval. I hope thats okay :)

Jonas-Sander:
> Haven't tried it out (no time currently) but we should add the steps to set up / run to the `CONTRIBUDING.md` in case there are some

I wouldn't do that yet, its not polished enough yet. In the future maybe.

PR #1442, opened 2024-04-10
Don't show average grade if grading system is not numerical.

Jonas-Sander:
Don't show the calculated average grade if the grading system is not numerical. To be able to do this I added `GradeResult.gradingSystem` and `GradeValue.gradingSystem`.  

Before:
image

After:
image
 
Requires #1441


nilsreichardt:
@Jonas-Sander There are few conflicts

nilsreichardt (approved):
LGTM

PR #1443, opened 2024-04-10
Add `--exclude` option to repo CLI

nilsreichardt:
Adds the option to exclude packages when running a command with the CLI. Example usage:

```sh
sz build_runner build --exclude holidays
```


nilsreichardt:
> What if --only foo and --exclude foo is passed?!?111 😁

Then will be `exclude` applied to the selection of packages that have been found using `--only`, e.g.

`--only holiydays,sharezone --exclude sharezone` will only run `holidays` :)

PR #1444, opened 2024-04-10
Remove `check-code-generation`

nilsreichardt:
We remove the `check-code-generation` check for now because we cannot get the following issue fixed in our CI: https://github.com/dart-lang/pub/issues/3314.


nilsreichardt:
@Jonas-Sander Feel free to merge when reviewed

Jonas-Sander (approved):

PR #1447, opened 2024-04-10
Add note about a school class license to the FAQ

nilsreichardt:
image


Jonas-Sander (approved):
Why not support@sharezone.net?

nilsreichardt:
> Why not [support@sharezone.net](mailto:support@sharezone.net)

Because I have configured some mailbox actions for plus@sharezone.net, e.g. they will be ranked as very important in my mailbox, never as spam, etc. (maybe in the future some other actions as well).

PR #1449, opened 2024-04-10
Add page to create a term

nilsreichardt:
https://github.com/SharezoneApp/sharezone-app/assets/24459435/8e1d5117-2f7b-4df8-b2bf-fca85163992c

('Aktuells Halbjahr' will be enabled by default - demo is old)


app/lib/grades/pages/create_term_page/create_term_page_controller.dart
```diff
@@ -0,0 +1,85 @@
...
+      throw InvalidTermNameException();
+    }
+
+    try {
+      final termId = TermId(Id.generate().id);
+      gradesService.addTerm(
+        id: termId,
+        name: view.name!,
+        isActiveTerm: view.isActiveTerm,
+        finalGradeType: const GradeTypeId('school-report-grade'),
```
Jonas-Sander, line 64:
```suggestion
        finalGradeType: GradeType.schoolReportGrade
```

Jonas-Sander (approved):
LGTM, in your video using a fullscreen dialog looks kinda overkill though :D

nilsreichardt:
> using a fullscreen dialog looks kinda overkill though

Yes true, but it's only a problem on desktop. Tablets and phones are taking half of the screen for the keyboard and then it's looks a bit better

PR #1450, opened 2024-04-10
Add more storage advantage to Sharezone Plus page

nilsreichardt:
image


Jonas-Sander (commented):
Why add this if it's not true?

Jonas-Sander (approved):

PR #1456, opened 2024-04-15
Complete 'Add Grade Dialog'

nilsreichardt:
## Description

This PR completes the grade dialog to enter a new grade.

The dialog has the following "smart" features:
* When the dialog is opened, the active term is selected.
* If the user selects a grading system that is not the grading system of the term, the Take Grade Into Account switch is disabled (and a note is displayed).
* When changing the term, the grading system of the newly selected term will be used (if the grade value is empty).
* When changing the grade type, the title will be set to the display name of the grade type if the title isn't the display name of the previous grade type.

## Demo

https://github.com/SharezoneApp/sharezone-app/assets/24459435/408dd30c-66d2-4fd5-ac54-352c4972fb89

https://github.com/SharezoneApp/sharezone-app/assets/24459435/652fc705-08b7-4db5-b72b-c06d820271d5

## Open questions

* When the user enters a grade (e.g. "1+") and changes the grade system (e.g. to "15 - 0 Punkte"), should we clear the grade value field?

## Things for later
* Add tests (the controller has so much logic, you can easily write ~50 tests for it) -> ideally we add this before we add editing grades
* Select the date from the past exams (connect exam feature with grades feature)


app/lib/grades/grades_service/grades_service.dart
```diff
@@ -494,7 +496,35 @@ enum PredefinedGradeTypes {
...
+      PredefinedGradeTypes.oralParticipation =>
+        const Icon(Symbols.record_voice_over, fill: 1),
+      PredefinedGradeTypes.vocabularyTest =>
+        const Icon(Symbols.text_rotation_none, fill: 1),
+      PredefinedGradeTypes.presentation =>
+        const Icon(Symbols.co_present, fill: 1),
+      PredefinedGradeTypes.other =>
+        const Icon(Symbols.other_admission, fill: 1),
+    };
+  }
```
nilsreichardt, line 527: Not sure if we want to change the location where we make this switch statement. Later, we will also load dynamically the icon for custom grade types (created by the user).

app/lib/grades/pages/grades_dialog/fields.dart
```diff
@@ -0,0 +1,553 @@
...
+    return MaxWidthConstraintBox(
+      maxWidth: 450,
+      child: SimpleDialog(
+        title: const Text("Notensystem auswählen"),
+        contentPadding: const EdgeInsets.fromLTRB(12, 12, 12, 16),
+        children: [
+          Padding(
+            padding: const EdgeInsets.fromLTRB(12, 0, 12, 12),
+            child: Text(
+              'Der erste Wert entspricht der besten Noten, z.B. bei dem Notensystem "1 - 6" ist "1" die beste Note.',
```
nilsreichardt, line 122: This isn't something that affects the calculations but will be important for the grading performance

Jonas-Sander:
> When the user enters a grade (e.g. "1+") and changes the grade system (e.g. to "15 - 0 Punkte"), should we clear the grade value field?

Yes

Jonas-Sander (approved):
Nice, LGTM 👍

app/lib/grades/grades_service/grades_service.dart
Jonas-Sander: Leave it there for now, we'll figure something out in the future

PR #1457, opened 2024-04-15
Add grading system to create term page

nilsreichardt:
image


app/lib/grades/pages/create_term_page/create_term_page.dart
```diff
@@ -136,6 +140,63 @@ class _NameField extends StatelessWidget {
...
+              for (final gradingSystem in GradingSystem.values)
+                ListTile(
+                  title: Text(gradingSystem.displayName),
+                  onTap: () {
+                    Navigator.of(context).pop<GradingSystem?>(gradingSystem);
+                  },
+                ),
+            ],
+          ),
+        );
```
nilsreichardt, line 168: Later, this dialog will be replaced with a shared one which is also used on grades dialog page (see #1456).

Jonas-Sander (approved):

PR #1463, opened 2024-04-19
Add grades subject page

nilsreichardt:
## Demo

https://github.com/SharezoneApp/sharezone-app/assets/24459435/89216b9f-84e4-4b2a-b2eb-934a9e369905

## Future tasks

* Add option switch between factors and percentage
* Add option to set weights per grade
* Add warning if sum of percentages isn't 100%


Jonas-Sander (approved):
Nice, LGTM 👍 thanks for the test

PR #1464, opened 2024-04-20
Add term settings page

nilsreichardt:
## Demo

https://github.com/SharezoneApp/sharezone-app/assets/24459435/4b785580-4ff3-469e-9a80-e0def2296863

## Description

This PR adds the settings page for terms.

For the subject weights: We merge all subjects and courses in the UI so it's more intuitive to the user what the setting does. Otherwise, we would only show the subjects for which the user has added a grade. If the user sets the weight for a subject that doesn't exist, we will automatically create one.


app/lib/grades/grades_service/grades_service.dart
```diff
@@ -640,6 +647,7 @@ class SubjectResult extends Equatable {
   final Design design;
   final IList<ConnectedCourse> connectedCourses;
   final GradeTypeId finalGradeTypeId;
+  final Weight weightingForTermGrade;
```
Jonas-Sander, line 650: Is this better for the UI or would you rather have `IMap<SubjectId, Weight> TermResult.subjectWeights` or sth similar?

Jonas-Sander (approved):
LGTM, I added this test: 'a subject will be added automatically to the term when setting the weight of a subject for a term'

PR #1476, opened 2024-04-21
Fix setting grade type weights for subject and term

nilsreichardt:
## Description

This was because I used `_getSubject()` and after calling `changeGradeTypeWeightForSubject()` Firestore wasn't quick enough to pull the update. Therefore, `_getSubject` returned the old subject. Same for the term.

## Demo

https://github.com/SharezoneApp/sharezone-app/assets/24459435/9687b3c0-51ad-4d4c-bb53-ee1e51716abc


Jonas-Sander (approved):
Can't you just listen to the stream and update the view for every new value? From what I'm seeing you keep a duplicate internal state via `_weights` currently, right?

nilsreichardt:
> From what I'm seeing you keep a duplicate internal state via _weights currently, right?

Yes, I'm duplicating the state. Yes, streaming would be better, but currently the settings page doesn't use streaming, so I will keep it like this but for the future streaming would better :+1:

PR #1479, opened 2024-04-21
Use `DateFormat` for grades date on term details page

nilsreichardt:
image


Jonas-Sander (approved):
Personally I like `21.04.2024` or `21.04.24` instead of `21.4.2024`, but do as you prefer

PR #1483, opened 2024-04-21
Add grades and plus page as notification navigation targets.

Jonas-Sander:
(I didn't try it out, since I can't set the action type with the Firebase UI it seems like)


nilsreichardt (approved):
LGTM

PR #1485, opened 2024-04-21
Remove `SubscriptionEnabledFlag` and `GradesEnabledFlag`.

Jonas-Sander:


nilsreichardt (approved):
LGTM

image

We need to keep in mind that the teachers/parents now also have the Sharezone Plus paywall

Jonas-Sander:
> LGTM
> image
> 
> We need to keep in mind that the teachers/parents now also have the Sharezone Plus paywall

yes, I'm gonna fix this in another PR

PR #1492, opened 2024-04-22
Fix displaying Sharezone Plus and the Grades feature

### LINKED ISSUE #1491: Fix: Loading on web when logged in as student will not show the grades, sharezone plus tiles in the drawer (only when updating the view by navigating to another section)

nilsreichardt:
Closes #1491


Jonas-Sander (approved):
Ah, nice, thanks!

PR #1501, opened 2024-04-22
Move privacy policy and TOS to new `legal` package.

Jonas-Sander:
Create `legal` package at `/lib/legal` for privacy policy and terms of service. This will enable us to use the same code to display the policies in our website.


nilsreichardt (approved):
LGTM

(I hope the CI doesn't break because of the Firebase upgrades in the `pubspec.lock`)

PR #1512, opened 2024-04-22
Make only 3 courses selectable if user doesn't have plus.

Jonas-Sander:
Automatically show only three Subjects if user doesn't have plus.

grafik


nilsreichardt (approved):
LGTM

PR #1556, opened 2024-04-24
Enable homework due date chips for stable track

nilsreichardt:
Just enable `showDueDateSelectionChips`. We later remove the unused code.


nilsreichardt:
Merging it without review since it's something we have on our Sharezone Plus sales page but disabled on the stable track. Pushing a new hotfix update.

Jonas-Sander (approved):

PR #1575, opened 2024-04-29
Fix: Deleting a term now sets `currentTerm` in Firestore doc to `null`.

### LINKED ISSUE #1573: When deleting all terms `currentTerm` is still set in the document.
When creating and then deleting a term the `currentTerm` field in the `Grades/AllInOne/` grades doc is still set.

Jonas-Sander:
When deleting all terms `currentTerm` would still be set to the old current term before. This is because if no current/active term was existing we didn't add the `currentTerm` key to the data map at all. This would leave the value unchanged. Now we explicitly set the field to null in this case.

Fixes #1573


nilsreichardt (approved):
LGTM

PR #1584, opened 2024-05-01
Fix share "let parents buy" option for iPad

nilsreichardt:


nilsreichardt:
Hot fix update, merging without a review

PR #1621, opened 2024-05-07
Make passing ids to `GradesService` optional, create `GradeInput`,`SubjectInput` classes

Jonas-Sander:
This PR enables the users of the `GradesService` to not need to generate Ids.

* Adds `SubjectInput` and `GradeInput` class (have no Id field)
* Make add methods return the Id of the newly created thing (term, subject, ...)
* `addCustomGradeType` no longer takes `GradeType`, but rather properties
* Make `Grade` class private since its only used for internal usage of `GradesService` now

### DIFF
--- a/app/lib/grades/grades_service/grades_service.dart
+++ b/app/lib/grades/grades_service/grades_service.dart
@@ -301,16 +301,16 @@ class GradesService {
     return gradeId;
   }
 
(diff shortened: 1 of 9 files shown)


nilsreichardt (approved):
LGTM

PR #1644, opened 2024-05-22
Rework homework page for teacher and parents

Jonas-Sander:
This reworks the homework page for teachers and parents in a way that is very similar to the current student page.

Small difference: There is no overflow button in the bottom action bar (overflow is used by students where they can select "mark all overdue homeworks as completed")

I also added the `firebase_hausaufgabenheft_logik` into the `hausaufgabenheft_logik`, since I don't really see that much benefit to separate it.

| Parents | Teacher |
|--------|--------|
| image | image |
| image |  image |


nilsreichardt (approved):
Nice work! I don't have the time to review it. Just review it by yourself 👍

PR #1668, opened 2024-06-11
Fix typo in onboarding

nilsreichardt:


nilsreichardt:
Skipping the review because the new update should be pushed now

PR #1669, opened 2024-06-11
Bump app version to v2.1.0

nilsreichardt:


nilsreichardt:
Skipping the review because the new update should be pushed now

PR #1691, opened 2024-07-25
Use adaptive platform density for privacy policy.

Jonas-Sander:
With 219bea1093928fa3192b99a215eb5c6e1f53fb7e we started to use `VisualDensitySetting.standard()` on all platforms instead of `VisualDensitySetting.adaptivePlatformDensity()`. Since the latter is better for the privacy policy UI on desktops we override it on this page.

| Before | After |
| - | - |
| grafik | grafik |

### DIFF
--- a/app/lib/legal/privacy_policy/privacy_policy_page.dart
+++ b/app/lib/legal/privacy_policy/privacy_policy_page.dart
@@ -54,14 +54,14 @@ class PrivacyPolicyPage extends StatelessWidget {
                     Provider.of<ThemeSettings>(context, listen: false);
                 final newSettings = _createPrivacyPolicyThemeSettings(
                     context, themeSettings, config);
-                // As of writing this we always use
+                // As of this writing, we always use
                 // VisualDensitySetting.standard() instead of
                 // VisualDensitySetting.adaptivePlatformDensity() on all
-                // platforms. This might sense for the general UI, but
-                // for the privacy policy page it is way better use the adaptive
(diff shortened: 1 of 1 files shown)


app/lib/legal/privacy_policy/privacy_policy_page.dart
```diff
@@ -52,8 +52,19 @@ class PrivacyPolicyPage extends StatelessWidget {
...
+                final newSettings = _createPrivacyPolicyThemeSettings(
                     context, themeSettings, config);
+                // As of writing this we always use
+                // VisualDensitySetting.standard() instead of
+                // VisualDensitySetting.adaptivePlatformDensity() on all
+                // platforms. This might sense for the general UI, but
+                // for the privacy policy page it is way better use the adaptive
+                // spacing. This will make the TOC tiles to be more compact for
+                // desktop (since users will most probably use a mouse there)
+                // and more spaced out for users using touch.
```
nilsreichardt, line 64: Suggestion (feel free to not use it):

```suggestion
                // As of this writing, we always use
                // VisualDensitySetting.standard() instead of
                // VisualDensitySetting.adaptivePlatformDensity() on all
                // platforms. This may make sense for the general UI, but for
                // the privacy policy page it is much better to use the adaptive
                // spacing. This will make the TOC tiles more compact for
                // desktop (since users will most likely be using a mouse there)
                // and more spread out for touch users.
```

nilsreichardt (approved):
LGTM

PR #1700, opened 2024-08-10
Remove Sharezone Wrapped

Jonas-Sander:
Since few users opened Sharezone Wrapped in the end, we want to remove the code for now, as we are unsure if we're going to be using it next year. If so, we can re-add it anyways.


nilsreichardt (approved):
LGTM

PR #1712, opened 2024-08-16
Enable changing display type of weights (factor or percentage)

Jonas-Sander:
Adds changing the display type of weights (as factor or as percent) in the term settings.

[Changing weight display type.webm](https://github.com/user-attachments/assets/131389f5-771e-4451-ae5b-fac602811370)

| Before | After |
|--------|--------|
| before | after |

The numbers are currently rounded in the weight overview, I'll fix this in another PR.


Sharezone-Admin (approved):
(Jonas S here, approving with this account so that I can merge via merge queue)

Jonas-Sander:
Nvm, would need to add the account to code owners. Merging with override.

PR #1703, opened 2024-08-11
Upgrade to Flutter v3.24.2

nilsreichardt:
Migrations:
* [Generic types in PopScope](https://docs.flutter.dev/release/breaking-changes/popscope-with-result)
* [Deprecate `ButtonBar` in favor of `OverflowBar`](https://docs.flutter.dev/release/breaking-changes/deprecate-buttonbar)

PR #1726, opened 2024-09-04
Use fvm in devenv again.

Jonas-Sander:
Needed for #1703 (or upgrading flutter in general right now), because using the current nix version of flutter 3.24 lead me to this bug: https://github.com/flutter/flutter/issues/153238 and I don't want to block upgrading flutter further, since until the nix version of flutter is upgraded it will probably take a week or more.

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -9,17 +9,6 @@
 { pkgs, inputs, lib, ... }:
 
 let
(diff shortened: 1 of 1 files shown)


nilsreichardt (approved):
LGTM

PR #1751, opened 2024-10-10
Fix "Support kontaktieren" button on our website

nilsreichardt:


website/lib/support_page.dart
```diff
@@ -15,8 +15,6 @@ import 'package:url_launcher/url_launcher.dart';
 
 import 'widgets/section.dart';
 
-const phoneNumber = '+49 1516 7754541';
```
nilsreichardt, line 18: Also removing the phone number since it's not used on the page

Jonas-Sander (approved):

PR #1773, opened 2024-10-16
Only show ad info dialog for free users

nilsreichardt:
This PR fixes that we showed the dialog also to plus users. To fix this, we listen to the `shouldShowInfoDialog`. It's better to transform this to a listenable value because the `SubscriptionService` will need a short moment to load the user document.

### DIFF
--- a/app/lib/dashboard/dashboard_page.dart
+++ b/app/lib/dashboard/dashboard_page.dart
@@ -123,6 +123,8 @@ class _AdsInfoDialogListener extends StatefulWidget {
 }
 
 class _AdsInfoDialogListenerState extends State<_AdsInfoDialogListener> {
+  // The AdsController already has a flag for this, but just to be safe for race
+  // conditions, we also keep track of it here.
   bool hasShownDialog = false;
 
   @override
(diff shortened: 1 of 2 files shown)


app/lib/dashboard/dashboard_page.dart
```diff
@@ -88,40 +88,60 @@ class _DashboardPageState extends State<DashboardPage> {
...
+  Widget build(BuildContext context) {
+    final showDialog = context.watch<AdsController>().shouldShowInfoDialog;
+    if (showDialog && !hasShownDialog) {
+      hasShownDialog = true;
+      context.read<AdsController>().shouldShowInfoDialog = false;
+
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        showAdInfoDialog(context);
+      });
+    }
```
nilsreichardt, line 140: Not the cleanest solution to call a dialog in the `build` method, but I don't had a better/easier idea. Open for suggestions

Jonas-Sander: Don't worry, It's fine like this

Jonas-Sander (approved):

PR #1809, opened 2025-02-04
Add error in testing code when adding term with subject with no grades to prevent developer confusion.

Jonas-Sander:


nilsreichardt (approved):
LGTM

PR #1811, opened 2025-02-04
Add error for developers in grade system tests.

Jonas-Sander:
Adds `bool throwErrorForSubjectsWithNoGrades` to `GradesTestController.createTerm`. This is to prevent developers from being confused when they create a term with a subject that has no grades and then try to access the subject of the term. This would throw an error as the term won't have the subject added, as it has no grades for it (yet).

Example for confusing behavior:
```
testController.createTerm(termWith(id: TermId('foo'), subjects: [
  subjectWith(id: SubjectId('maths')),
]));
// Throws error as the term does not have the 'maths' subject, because 
// it was not added to the term because there is no grade for 'maths' fo
// this term.
testController.term(TermId('foo')).subject(SubjectId('maths')).name;
```

Now there would be an error thrown when calling `testController.createTerm` which makes debugging easier.

### DIFF
--- a/app/test/grades/grade_types_test.dart
+++ b/app/test/grades/grade_types_test.dart
@@ -329,6 +329,7 @@ void main() {
           id: const TermId('foo'),
           subjects: [subjectWith(id: const SubjectId('bar'))],
         ),
+        throwErrorForSubjectsWithNoGrades: false,
       );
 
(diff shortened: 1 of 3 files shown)


Jonas-Sander:
Forgot that I already created #1809 xD

PR #1815, opened 2025-02-08
Update to Flutter 3.27.4

Jonas-Sander:
Update to Flutter 3.27.4

Notes:
Use of pub workspaces (should result in faster analyzer)

https://docs.flutter.dev/release/breaking-changes/wide-gamut-framework
We shouldn't use some deprecated `Color` attributes, but we use these in database and some old code that needs to be changed first to accommodate these changes. For now I'd just ignore this and fix this later (when they are removed in the future we're forced to do this anyways).


nilsreichardt:
Thanks for the migration! ❤️ We just need to keep in mind that with 3.27 there are some changes to dynamic / deep links. Maybe they also need some migration. Alternatively, we can also just break them (and don't publish an update), since we need to replace Firebase Dynamic Links in the next weeks anyway.

Jonas-Sander:
> since we need to replace Firebase Dynamic Links in the next weeks anyway.

Do you already have a plan in mind how we do this?

Jonas-Sander:
> Thanks for the migration! ❤️ We just need to keep in mind that with 3.27 there are some changes to dynamic / deep links. Maybe they also need some migration. Alternatively, we can also just break them (and don't publish an update), since we need to replace Firebase Dynamic Links in the next weeks anyway.

I would skip this for now and merge so that we have the faster analyzer speed because of pub workspaces

Jonas-Sander:
@nilsreichardt Could you regenerate the GitHub PAT that is needed for device preview? I guess this was probably yours

nilsreichardt (approved):
(Couldn't review, no time because of exams 😢)

nilsreichardt:
> @nilsreichardt Could you regenerate the GitHub PAT that is needed for device preview? I guess this was probably yours

Re-generate a new PAT with @Sharezone-Admin account 👍

PR #1816, opened 2025-02-08
Rename and update command to replace local golden files with those from GitHub Actions (`sz replace-goldens`).

Jonas-Sander:
Renames `pick-codemagic-goldens` to `replace-goldens` and makes the command automagically replace the local goldens with the one from GitHub (one has to download them first though).

To use:
1. Download failed goldens `.zip` from the GitHub Actions Job.
2. Extract the folders somewhere into this repo
3. Run `sz replace-goldens`.


nilsreichardt (approved):
LGTM

PR #1836, opened 2025-03-19
Add copy user id to docs

nilsreichardt:
image


nilsreichardt:
(Merging without review since I need this link in a customer email now)

Issue #1879, opened 2025-06-11
Deployment for our macOS app doesn't work anymore


nilsreichardt:
```
ttribute's length must be >= 1 and <= 64, but it was 67
  for name_attribute in self.certificate.subject
productsign: error: Could not find appropriate signing identity for “”.
Running "keychain use-login" in /Users/runner/work/sharezone-app/sharezone-app.
Use login keychain /Users/runner/Library/Keychains/login.keychain-db as system default keychain
Set keychain /Users/runner/Library/Keychains/login.keychain-db to system default keychain
Unhandled exception:
ProcessRunnerException: Running "bash -c APP_NAME=$(find $(pwd) -name "*.app") && 
PACKAGE_NAME=$(basename "$APP_NAME" .app).pkg && 
xcrun productbuild --component "$APP_NAME" /Applications/ unsigned.pkg && 
INSTALLER_CERT_NAME=$(keychain list-certificates | jq -r '.[] | select(.common_name | contains("Mac Developer Installer")) | .common_name' | head -1) && 
xcrun productsign --sign "$INSTALLER_CERT_NAME" unsigned.pkg "$PACKAGE_NAME" && 
rm -f unsigned.pkg" in /Users/runner/work/sharezone-app/sharezone-app/app exited with code 1
productbuild: Adding component at /Users/runner/work/sharezone-app/sharezone-app/app/build/macos/Build/Products/Release-prod/Sharezone.app
productbuild: Wrote product to unsigned.pkg
productbuild: Supported OS versions: [Min: 10.15, Before: None]
List available code signing certificates in keychain /Users/runner/Library/codemagic-cli-tools/keychains/10-06-25_yzfn0t0w.keychain-db
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/codemagic/models/certificate.py:99: UserWarning: Attribute's length must be >= 1 and <= 64, but it was 67
  for name_attribute in self.certificate.subject
productsign: error: Could not find appropriate signing identity for “”.
:
List available code signing certificates in keychain /Users/runner/Library/codemagic-cli-tools/keychains/10-06-25_yzfn0t0w.keychain-db
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/codemagic/models/certificate.py:99: UserWarning: Attribute's length must be >= 1 and <= 64, but it was 67
  for name_attribute in self.certificate.subject
[...]

PR #1896, opened 2025-07-01
Try to fix macOS builds

nilsreichardt:


nilsreichardt:
Fixed with https://github.com/SharezoneApp/sharezone-app/issues/1879#issuecomment-3026906101

PR #1911, opened 2025-08-24
Fix `devenv` to work with Android emulator.

Jonas-Sander:
[Devenv](https://devenv.sh/) is way to make the development environment reproducible using Nix.

Before the whole set up was only working with the browser for me. Now I can also use the emulator that can be created by running the `create-emulator` script (only when using devenv). This setup will probably only work for NixOS (maybe Windows with WSL?) but not for macOS using devenv yet.

Had to let `.vscode/settings.json` now be ignored by git  because I don't use `fvm`. When using `devenv` to set up my environment the `settings.json` flutter path pointing to the fvm folder breaks flutter for VS Code.   
That shouldn't be a problem since FVM [automatically prompts/makes the changes](https://fvm.app/documentation/guides/vscode#automatic-detection-and-configuration) in VS Code to add the flutter path in `settings.json`.

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -31,6 +31,7 @@ in
     pkgs.unzip
 
     # For Flutter Web development
+    # pkgs.google-chrome lead to long local builds that don't happen with pkgs.chromium.
     pkgs.chromium
 
     # Used by sharezone developers / sz cli


nilsreichardt (approved):
LGTM

PR #1914, opened 2025-08-24
Replace dynamic links with`app_links` package

Jonas-Sander:
Use [`app_links`](https://pub.dev/packages/app_links) package since dynamic links is deprecated.

As a quick fix we generate the joinLinks locally since regenerating all links in the backend will take too long until the service is deprecated.

- [x] Android
- [x] iOS 
- [ ] macOS (should work in theory but didn't when testing locally)
- [ ] Web (?)

On macOS in the `app_links` docs it only says to add something to the `Info.plist` file, but the AI says that it has to be added as a "Associated Domains" capability, as the `Info.plist` is only for schemes, not `https`-URLs. I've added the domain under "Associated Domains" like for iOS and not added anything to the plist file.


Jonas-Sander:
@nilsreichardt It seems to work for iOS and Android, couldn't get macOS working though. I think I'd just like to merge this anyways, what do you think?

lib/group_domain_models/lib/src/models/group_info.dart
```diff
@@ -13,7 +13,16 @@ class GroupInfo {
...
+  final String? sharecode;
+  // Originally we used a joinLink from the backend, but since Firebase Dynamic
+  // Links was deprecated, we needed a quick solution for new links.
+  // Regenerating all existing join links in the database is not feasible right
+  // now, so we compute the joinLink from the sharecode (before we used another
+  // link format).
+  String? get joinLink =>
+      sharecode == null
+          ? null
+          : "https://web.sharezone.net/link?type=joinbykey&data=$sharecode";
```
nilsreichardt, line 25: I think a positive side effect of this is that the `joinLink` uses the personal Sharecode now, right? IIRC, used everyone the same `joinLink`

nilsreichardt (approved):
LGTM (haven't tested it). Should I test it on iOS with a real device?

Jonas-Sander:
> LGTM (haven't tested it). Should I test it on iOS with a real device?

Yes please

nilsreichardt:
✅ App open, signed in:

https://github.com/user-attachments/assets/5119fe01-0cb7-4fab-96df-657d82b419fd

❌ App not open, signed in:

https://github.com/user-attachments/assets/31e361d9-2143-4365-9bba-22fc05ac613e

Multiple tries to join the course

❌ App open, registration

https://github.com/user-attachments/assets/5babed20-d35c-456d-8844-534f0d010d47

Multiple tries to join the course

❌ App not open, registration

https://github.com/user-attachments/assets/feb3e2a4-767f-4741-a46f-b5d09a027b91

Multiple tries to join the course

Jonas-Sander:
Thanks, I'm gonna go ahead and merge this then. The registration behavior can be enhanced later.

nilsreichardt:
https://github.com/user-attachments/assets/69886992-72f9-4ad0-845e-2c08ebf8eda9

Also works on macOS 🥳

PR #1916, opened 2025-08-25
Host `web.sharezone.net/.well-known` files ourselves

Jonas-Sander:
I think these files have been automatically hosted by dynamic links so far, but as the service is now deprecated we have to host them ourselves. Since I couldn't get .gitignore to only not ignore files in `app/build/web/.well-known`, I created the folder `app/web_static_root` which is the folder that gets copied into the `app/build/web` (root) folder.


Jonas-Sander:
@nilsreichardt I'm gonna go ahead and merge since this is needed to even get #1914 testable on iOS and macOS.

PR #1937, opened 2025-09-19
Fix opening links in feedback chat

nilsreichardt:
You couldn't open the links sent in the feedback chat.


Jonas-Sander (commented):
If we're security conscious, we should maybe add a pop-up that displays the URL and a button to continue opening it.

That's because people could write a link like this: [https://www.google.com](https://youtu.be/dQw4w9WgXcQ) which actually leads to another destination (phishing, malware, etc.) 

(Or some other way which balances security with usability so one doesn't need to click another button always)

WDYT?

PR #1966, opened 2025-10-11
Add option to delete grade subjects

### LINKED ISSUE #1529: Allow users to delete Subjects

nilsreichardt:
https://github.com/user-attachments/assets/123d314e-370c-4edd-a36a-e5a66820ef47

Closes #1529

### DIFF
--- a/app/test/grades/pages/subjects_page/subjects_page_test.dart
+++ b/app/test/grades/pages/subjects_page/subjects_page_test.dart
@@ -6,6 +6,7 @@
 //
 // SPDX-License-Identifier: EUPL-1.2
 
+import 'package:common_domain_models/common_domain_models.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:group_domain_models/group_domain_models.dart';
@@ -13,9 +14,8 @@ import 'package:provider/provider.dart';
 import 'package:sharezone/grades/grades_service/grades_service.dart';
 import 'package:sharezone/grades/pages/subjects_page/subjects_page.dart';
 import 'package:sharezone/grades/pages/subjects_page/subjects_page_controller_factory.dart';
-import 'package:sharezone_widgets/sharezone_widgets.dart';
-import 'package:common_domain_models/common_domain_models.dart';
 import 'package:sharezone_localizations/sharezone_localizations.dart';
+import 'package:sharezone_widgets/sharezone_widgets.dart';
 
 import '../../grades_test_common.dart';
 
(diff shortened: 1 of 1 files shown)


app/lib/grades/pages/subjects_page/subjects_page.dart
```diff
@@ -0,0 +1,307 @@
...
+Das hat noch einen Vorteil: Du kannst deine Noten nach Fächern sortieren und später deine Entwicklung in einem Fach über mehrere Jahre hinweg verfolgen (diese Funktion ist bald verfügbar).
+
+Sharezone legt automatisch ein Notenfach an, sobald du eine Note in einem Kurs erstellst.
+''',
+      ),
+      backgroundColor: Theme.of(
+        context,
+      ).colorScheme.primary.withValues(alpha: 0.1),
+    );
+  }
```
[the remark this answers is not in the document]
Jonas-Sander: Very good! :) 

I think it is a really helpful idea to include this help text.

app/lib/grades/pages/subjects_page/subjects_page_controller.dart
```diff
@@ -0,0 +1,251 @@
...
+      acc,
+      gradeType,
+    ) {
+      final name =
+          gradeType.displayName ??
+          gradeType.predefinedType?.toUiString() ??
+          'Unbekannt';
+      acc[gradeType.id] = name;
+      return acc;
+    });
```
Jonas-Sander: I found this very hard to understand right away because we don't use fold very often and I'm not sure if its the best fit here. Overall with the usage later this kind of feels like a code smell because its seems too complicated.  

I think having to use:
```dart
          gradeType.displayName ??
          gradeType.predefinedType?.toUiString() ??
          'Unbekannt';
```
is a sign of bad design of my GradeType code, because I think we probably need this somewhere else as well (duplication).  

Anyways, as a start can't we use something like this to make it a bit simpler?
```dart
    final gradeTypeNames = gradeTypes.map((gradeType) {
      final gradeTypeDisplayName =
          gradeType.displayName ??
          gradeType.predefinedType?.toUiString() ??
          'Unbekannt';

      return MapEntry(gradeType.id, gradeTypeDisplayName);
    }).toMap();
```

app/test/grades/grades_service_test.dart
```diff
@@ -40,5 +42,75 @@ void main() {
         isNull,
       );
     });
+
+    test('deleteSubject removes subject and grades from all terms', () {
```
Jonas-Sander: If you're testing for terms and grades (plural) then you should actually test it with several terms and grades IMO. Maybe 2 terms, one with 2 grades from the subject ot be deleted and one with 2 grades, 1 to be deleted one from another subject. In the end make sure the grades that should be deleted are deleted and the grade that should not be deleted is not deleted.

(One could argue that one should separate this into two tests (deleting everything it should, not deleting anything it should not) but that's kleinscheißerei, I don't care :D)

PR #1969, opened 2025-10-15
Add `prefer_final_locals` to `sharezone_lints`

nilsreichardt:
**BAD**
```dart
void badMethod() {
  var label = 'hola mundo! badMethod'; // LINT
  print(label);
}
```

**GOOD:**
```dart
void goodMethod() {
  final label = 'hola mundo! goodMethod';
  print(label);
}
```

WDYT? Makes it the code more readable? I have no strong opinion here.


nilsreichardt:
Then I would close it, just saw in one method and thought maybe it's a good thing but after running the `dart fix --apply` it was more changes than I thought.

PR #1987, opened 2025-12-07
Disable Gemini Code Review summaries

nilsreichardt:
WDYT? I like the code suggestions but the summary in most cases useless.


Jonas-Sander:
Yeah let's disable it. 

Why did Codex give a thumbs up? 😂

